Issue #2, opened 2019-02-13
Initial population from edgex-go


tsconn23:
The initial population of this repo should contain code from the /pkg folder of edgex-go. This module will eventually replace said folder.

PR #3, opened 2019-02-13
Initial code commit from edgex-go/pkg

tsconn23:
#2 

A couple things to note:

- Legacy models and client related to metadata/scheduler have been preserved. IDs in these models have been changed to string from BSON. If we are going to preserve these for Edinburgh after all, then work needs to be done in edgex-go
- Service keys that were previously in `internal` have been created here as constants. Repoint in `edgex-go` when consuming this module.


tsconn23:
recheck

jduranf:
Why not change `models` by `contract`?

main.go
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package contracts
+
+//Must be a .go file of some sort in the root, otherwise "make test" will fail.
```
lenny-intel: I didn't need this in mine. Are you able to run make test in https://github.com/lenny-intel/go-mod-registry/tree/InitialImpl?

tsconn23: @lenny-intel But you do have a go file in your root -- interface.go -- which is why it works.

ernestojeda:
recheck

main.go
lenny-intel: aHH, right. I was thinking main.go specifically, not any .go file. Works for me.

tsconn23:
> Why not change `models` by `contract`?

@jduranf Well, my thinking was the repo name scopes the models as the core service contract. Plus, leaving the package names the same makes it easier to find/replace the imports when it comes time to consume this in edgex-go. In edgex-go we already alias these as `contract` anyway.

Issue #11, opened 2019-02-22
Add Version File for Tagging Support


tsconn23:
Add a version file for tagging support. Initial rev @ v0.0.1

PR #13, opened 2019-02-22
Adding VERSION file for tagging

tsconn23:
#11 


VERSION
```diff
@@ -0,0 +1 @@
+0.0.1
```
lenny-intel: Do we need the "v", i.e. v0.0.1, as I did in the go-mod-registry or doesn't matter?

tsconn23: @lenny-intel Yeah, I had this same question. I will add it, and I suspect we'll need to add to edgex-go as well.

Issue #14, opened 2019-02-24
Add AutoEvents to Device model


cloudxxx8:
The purpose of this change is to provide a new mechanism for device services to autonomously generate events, and to be able to take advantage of devices which can push new readings rather than relying on polling.

tsconn23:
Specification of AutoEvents field taken from the following issue
https://github.com/edgexfoundry/edgex-go/issues/946

```
We propose that an "AutoEvents" field is added to "Device". This will contain 0..n entries which are defined as { resource, interval, onChange }.
resource : the name of the resource in the device profile which describes the event to generate.
interval : for devices which must be polled, the time between queries. For devices which can push, the minimum interval between generated events (ie, a rate limit).
onChange : boolean. If true, only generate events when readings change.
```

PR #15, opened 2019-02-25
Device support for multiple protocols, auto events

tsconn23:
Issues #7 and #14 

NOTE: I went with the property name AutoEvent.Frequency as opposed to AutoEvent.Interval because the new Scheduler contract has an `Interval` type. I thought that using the name Interval here would imply that it was of type `Interval` which is not the case.

### DIFF
--- a/models/autoevent.go
+++ b/models/autoevent.go
@@ -1,0 +1,34 @@
+/*******************************************************************************
+ * Copyright 2019 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package models
+
+import "encoding/json"
+
+type AutoEvent struct {
+	Frequency int32  `json:"frequency,omitempty"`
+	OnChange  bool   `json:"onChange,omitempty"`
+	Resource  string `json:"resource,omitempty"`
+}
+
+/*
+ * String function for representing an auto-generated event from a device.
+ */
+func (a AutoEvent) String() string {
+	out, err := json.Marshal(a)
+	if err != nil {
+		return err.Error()
+	}
+	return string(out)
+}
--- a/models/device.go
+++ b/models/device.go
@@ -26,45 +26,48 @@ import (
  */
 type Device struct {
 	DescribedObject
-	Id             string         `json:"id"`
-	Name           string         `json:"name"`           // Unique name for identifying a device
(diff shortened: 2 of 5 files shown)


models/device.go
```diff
@@ -26,45 +26,48 @@ import (
...
-	LastReported   int64          `json:"lastReported"`   // Time (milliseconds) that the device reported data to the core microservice
-	Labels         []string       `json:"labels"`         // Other labels applied to the device to help with searching
-	Location       interface{}    `json:"location"`       // Device service specific location (interface{} is an empty interface so it can be anything)
-	Service        DeviceService  `json:"service"`        // Associated Device Service - One per device
-	Profile        DeviceProfile  `json:"profile"`        // Associated Device Profile - Describes the device
+	Id             string                       `json:"id"`
+	Name           string                       `json:"name"`           // Unique name for identifying a device
+	AdminState     AdminState                   `json:"adminState"`     // Admin state (locked/unlocked)
+	OperatingState OperatingState               `json:"operatingState"` // Operating state (enabled/disabled)
+	Protocols      map[string]map[string]string `json:"protocols"`      // A map of supported protocols for the given device
```
iain-anderson: Can we typedef ProtocolProperties as map[string]string, and make the type for Protocols be map[string]ProtocolProperties

models/autoevent.go
```diff
@@ -0,0 +1,34 @@
...
+import "encoding/json"
+
+type AutoEvent struct {
+	Frequency int32  `json:"frequency,omitempty"`
+	OnChange  bool   `json:"onChange,omitempty"`
+	Resource  string `json:"resource,omitempty"`
+}
+
+/*
+ * String function for representing a device
```
iain-anderson: "an AutoEvent"

models/device.go
tsconn23: No because as discussed on the call yesterday, this would require ProtocolProperties to be a struct type -- which is the same as we have right now with Addressable. This is considered too restrictive for the desired solution since a struct has a fixed list of properties that aren't flexible enough to support all cases. If it were, we'd make a map[string]Addressable and call it a day.

models/autoevent.go
tsconn23: Commented updated

models/device.go
```diff
@@ -26,45 +26,48 @@ import (
...
-	AdminState     AdminState     `json:"adminState"`     // Admin state (locked/unlocked)
-	OperatingState OperatingState `json:"operatingState"` // Operating state (enabled/disabled)
-	Addressable    Addressable    `json:"addressable"`    // Addressable for the device - stores information about it's address
-	LastConnected  int64          `json:"lastConnected"`  // Time (milliseconds) that the device last provided any feedback or responded to any request
-	LastReported   int64          `json:"lastReported"`   // Time (milliseconds) that the device reported data to the core microservice
-	Labels         []string       `json:"labels"`         // Other labels applied to the device to help with searching
-	Location       interface{}    `json:"location"`       // Device service specific location (interface{} is an empty interface so it can be anything)
-	Service        DeviceService  `json:"service"`        // Associated Device Service - One per device
-	Profile        DeviceProfile  `json:"profile"`        // Associated Device Profile - Describes the device
+	Id             string                       `json:"id"`
```
tonyespy: What's with all these other fields being changed? Whitespace cleanup?

tsconn23: @tonyespy `go fmt`

PR #18, opened 2019-02-28
Updated general client and service to fetch health.

brandonforster:
This PR is required for the PR for the following issue in the monorepo:

https://github.com/edgexfoundry/edgex-go/issues/833

### DIFF
--- a/clients/constants.go
+++ b/clients/constants.go
@@ -29,6 +29,7 @@ const (
 	ApiDeviceProfileRoute      = "/api/v1/deviceprofile"
 	ApiDeviceServiceRoute      = "/api/v1/deviceservice"
 	ApiEventRoute              = "/api/v1/event"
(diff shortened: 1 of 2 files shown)


brandonforster:
> Two things
> 
> * The `consulUrl` parameter should be renamed to `registryUrl` since Consul indicates a specific implementation
> * I think it is more appropriate for this to be obtained from the `go-mod-registry` module since the GeneralClient is used to make calls to EdgeX services, not infra.
> 
> If you look in the ReadMe for the [go-mod-registry](https://github.com/edgexfoundry/go-mod-registry) you will see one of its capabilities is "Check the health status of dependent services via the Registry service." Thus the capability you're looking for appears to already exist and doesn't need to be added here.

Thanks. I didn't realize this was a thing. I'll close this and the other PR, regroup and resubmit.

PR #25, opened 2019-03-09
Add godoc to AutoEvent and correct frequency type

### LINKED ISSUE #24: Add godoc to AutoEvent and correct the type of frequency
The type of frequency ought to be uint32 rather than int32
https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/autoevent.go#L20

Also, we had better add some document into autoevent.go

cloudxxx8:
fix #24
Signed-off-by: Cloud Tsai <cloudxxx8@gmail.com>

### DIFF
--- a/models/autoevent.go
+++ b/models/autoevent.go
@@ -16,10 +16,18 @@ package models
 
 import "encoding/json"
 
+// AutoEvent supports auto-generated events sourced from a device service
 type AutoEvent struct {
-	Frequency int32  `json:"frequency,omitempty"`
-	OnChange  bool   `json:"onChange,omitempty"`
-	Resource  string `json:"resource,omitempty"`
+	// Frequency indicates how often the specific resource needs to be polled.
+	// It represents the number of events per day.
+	// For example, 86400 would be once per second, 24 would be once per hour etc.
+	Frequency uint32 `json:"frequency,omitempty"`
+	// OnChange indicates whether the device service will generate an event only,
+	// if the reading value is different from the previous one.
+	// If true, only generate events when readings change
+	OnChange bool `json:"onChange,omitempty"`
+	// Resource indicates the name of the resource in the device profile which describes the event to generate
+	Resource string `json:"resource,omitempty"`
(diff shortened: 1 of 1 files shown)


cloudxxx8:
@iain-anderson please help review and correct my wording.  thanks

models/autoevent.go
```diff
@@ -16,10 +16,21 @@ package models
 
 import "encoding/json"
 
+/*
+ * This file is the model for the AutoEvent object in EdgeX
```
tsconn23: Maybe instead "`AutoEvent supports auto-generated events sourced from a device service`"

models/autoevent.go
```diff
@@ -16,10 +16,21 @@ package models
...
+ *
+ * AutoEvent struct
+ */
 type AutoEvent struct {
-	Frequency int32  `json:"frequency,omitempty"`
-	OnChange  bool   `json:"onChange,omitempty"`
-	Resource  string `json:"resource,omitempty"`
+	// Frequency indicates that how often the specific resource need to be polled.
```
tsconn23: "`Frequency indicates how often the specific resource needs to be polled.`"

models/autoevent.go
```diff
@@ -16,10 +16,21 @@ package models
...
+	// It represents the number of events per day.
+	// For example, 86400 would be once per second, 24 would be onece per hour etc.
```
tsconn23: "once" misspelled as "onece"

models/autoevent.go
```diff
@@ -16,10 +16,21 @@ package models
...
+	Frequency uint32 `json:"frequency,omitempty"`
+	// OnChange indicates whether generate the events only if the reading value is different from the previous one.
```
tsconn23: "`OnChange indicates whether the device service will generate an event only if the reading value is different from the previous one.`"

cloudxxx8:
@tsconn23 thanks for the suggestion, I've corrected the according to your comment

Issue #20, opened 2019-03-05
Remove Schedule / ScheduleEvent models


tsconn23:
We need to remove the Schedule and ScheduleEvent models from this project as their supported endpoint is deprecated. Remove associated tests as well.

jduranf:
Done via #28

PR #28, opened 2019-03-11
Removal of Schedule/ScheduleEvent

tsconn23:
#20 

NOTE: DeviceReports had a link to ScheduleEvent. I have replaced this for now with IntervalAction. However, after talking with @jpwhitemn , it appears DeviceReports are unused and may also be removed from the system.

Once this PR is merged, I will create the PR in edgex-go for removal of the handling logic/tests.


jduranf:
If I'm not wrong, `ApiScheduleRoute` and `ApiScheduleEventRoute`  in [constants.go](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/clients/constants.go#L42) can be removed.
And I think `Event` field in [event.go](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/event.go#L34) can be also be removed. Maybe @cloudxxx8 can confirm this.

tsconn23:
> If I'm not wrong, `ApiScheduleRoute` and `ApiScheduleEventRoute` in [constants.go](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/clients/constants.go#L42) can be removed.
> And I think `Event` field in [event.go](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/event.go#L34) can be also be removed. Maybe @cloudxxx8 can confirm this.

Good catch. Removed the constants and I confirmed your comment about the `Event.Event` field so removed it. I'm thrilled to remove that stupid property BTW.

cloudxxx8:
Yes, I can confirm Event field in event.go could be removed as my investigation earlier.  It won't be used anymore.

Issue #32, opened 2019-03-14
Add Support For Binary Readings


tsconn23:
Per the Core WG call on 14-Mar-19 we need to start adding support for binary readings as part of the CBOR initiative for Edinburgh release. To that end, the following changes need to be made in the core contracts module.

- Add to ValueDescriptor model
  - "MimeType" property (string)
  - Modify unit tests accordingly
- Add to Reading model
  - "Blob" property ([]byte)
  - Modify unit tests accordingly
- Clients/coredata/event.go
  - The "Add" method should be refactored to look at the incoming context for a `Content-type` header. If found, then send the value through as part of the request. Otherwise default to JSON, which is the current behavior.

Please define a new [constant](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/clients/constants.go) for the new header value `application/cbor`. This will be usable by any application utilizing the contracts and will avoid naming mismatch issues.

lenny-intel:
I suggest the following Content Type constants to start with : 

- application/json
- application/cbor
- application/xml

tsconn23:
Yes, we already have a couple of those:
https://github.com/edgexfoundry/go-mod-core-contracts/blob/5d775d0480d8908efa18c30cac6dcb87cdc84e37/clients/request.go#L33

tonyespy:
> Per the Core WG call on 14-Mar-19 we need to start adding support for binary readings as part of the CBOR initiative for Edinburgh release. To that end, the following changes need to be made in the core contracts module.
> 
> * Add to ValueDescriptor model
>   
>   * "MimeType" property (string)
>   * Modify unit tests accordingly

I don't like MimeType here as it's a term usually associated with email attachments. I think ContentType is more appropriate, although to make it clear that it just applies binary readings, I suggested "BinaryContentType" in our meeting earlier today.

I also believe that this field needs to be added to PropertyValue, which is what is embedded in each deviceResource, as defined by the device profile. PropertyValue is used to populate the ValueDescriptor.

> * Add to Reading model
>   
>   * "Blob" property ([]byte)

Thoughts on maybe calling this BinaryValue?

>   * Modify unit tests accordingly
> * Clients/coredata/event.go
>   
>   * The "Add" method should be refactored to look at the incoming context for a `Content-type` header. If found, then send the value through as part of the request. Otherwise default to JSON, which is the current behavior.
> 
[...]

PR #35, opened 2019-03-15
Added CBOR data type constant and basic framework.

brandonforster:
Issue #32 

This should not be the final say, but is the initial refactoring required for work to continue.

### DIFF
--- a/clients/constants.go
+++ b/clients/constants.go
@@ -58,3 +58,10 @@ const (
 	SystemManagementAgentServiceKey = "edgex-sys-mgmt-agent"
 	SupportSchedulerServiceKey      = "edgex-support-scheduler"
 )
+
+const (
+	ContentType     = "Content-Type"
+	ContentTypeCBOR = "application/cbor"
+	ContentTypeJSON = "application/json"
+	ContentTypeYAML = "application/x-yaml"
+)
--- a/clients/metadata/device_profile.go
+++ b/clients/metadata/device_profile.go
@@ -126,7 +126,9 @@ func (dpc *DeviceProfileRestClient) Update(dp models.DeviceProfile, ctx context.
 }
 
 func (dpc *DeviceProfileRestClient) Upload(yamlString string, ctx context.Context) (string, error) {
-	return clients.PostRequest(dpc.url+"/upload", []byte(yamlString), clients.ContentYaml, ctx)
+	ctx = context.WithValue(ctx, clients.ContentType, clients.ContentTypeYAML)
+
+	return clients.PostRequest(dpc.url+"/upload", []byte(yamlString), ctx)
 }
 
 func (dpc *DeviceProfileRestClient) UploadFile(yamlFilePath string, ctx context.Context) (string, error) {
--- a/clients/request.go
+++ b/clients/request.go
@@ -30,12 +30,6 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 )
 
-const (
-	ContentType = "Content-Type"
-	ContentJson = "application/json"
-	ContentYaml = "application/x-yaml"
-)
-
 // Helper method to get the body from the response after making the request
 func getBody(resp *http.Response) ([]byte, error) {
 	body, err := ioutil.ReadAll(resp.Body)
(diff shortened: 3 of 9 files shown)


models/profileproperty.go
```diff
@@ -17,8 +17,9 @@ package models
 import "encoding/json"
 
 type ProfileProperty struct {
-	Value PropertyValue `json:"value"`
-	Units Units         `json:"units"`
+	Value     PropertyValue `json:"value"`
+	Units     Units         `json:"units"`
+	MediaType string        `json:"mediaType"`
```
tsconn23: Please put an additional `omitempty` tag on the json handling for MediaType

models/reading.go
```diff
@@ -34,6 +34,7 @@ type Reading struct {
 	Device   string `json:"device"`
 	Name     string `json:"name"`
 	Value    string `json:"value"` // Device sensor data value
+	Blob     []byte `json:"blob"` // Binary data payload
```
tsconn23: Per @tonyespy's [comment](https://github.com/edgexfoundry/go-mod-core-contracts/issues/32#issuecomment-473112193), we need to call this `BinaryValue` instead of `Blob`

models/reading_test.go
```diff
@@ -15,14 +15,17 @@
...
 	"reflect"
 	"strconv"
 	"testing"
 )
 
 var TestValueDescriptorName = "Temperature"
 var TestValue = "45"
-var TestReading = Reading{Pushed: 123, Created: 123, Origin: 123, Modified: 123, Device: TestDeviceName, Name: TestValueDescriptorName, Value: TestValue}
+var TestBlob = []byte{ 0xbf }
+var TestReading = Reading{Pushed: 123, Created: 123, Origin: 123, Modified: 123, Device: TestDeviceName, Name: TestValueDescriptorName, Value: TestValue, Blob: TestBlob}
```
tsconn23: Test will then need to be adjusted for `Blob` --> `BinaryValue` rename

clients/request.go
```diff
@@ -209,7 +210,7 @@ func PutRequest(url string, body []byte, ctx context.Context) (string, error) {
 	}
 
 	if body != nil {
-		req.Header.Set(ContentType, ContentJson)
+		req.Header.Set(ContentType, ContentTypeJson)
```
tsconn23: Check `Context` first to make sure ContentType isn't already set, then default to JSON. Same as `PostRequest()` above.

clients/metadata/device_profile.go
```diff
@@ -126,7 +126,7 @@ func (dpc *DeviceProfileRestClient) Update(dp models.DeviceProfile, ctx context.
 }
 
 func (dpc *DeviceProfileRestClient) Upload(yamlString string, ctx context.Context) (string, error) {
-	return clients.PostRequest(dpc.url+"/upload", []byte(yamlString), clients.ContentYaml, ctx)
+	return clients.PostRequest(dpc.url+"/upload", []byte(yamlString), ctx)
```
tsconn23: Notice that the previous implementation passed in the YAML content type. We've lost that with this change. Suggest you check the `Context` for `Content-Type`. If it's not there, then add it with the appropriate YAML value. The subsequent handling inside of `PostRequest` should then pick it up.

clients/request.go
brandonforster: Okay. So we're setting the content type to JSON in both the request header and the context, correct?

clients/metadata/device_profile.go
brandonforster: `fromContext` is still out of scope here, but it shouldn't be a problem to always set the `Content-Type` to YAML, right? We are uploading YAML right here, aren't we?

clients/constants.go
```diff
@@ -58,3 +58,10 @@ const (
 	SystemManagementAgentServiceKey = "edgex-sys-mgmt-agent"
 	SupportSchedulerServiceKey      = "edgex-support-scheduler"
 )
+
+const (
+	ContentType       = "Content-Type"
+	ContentTypeBinary = "application/cbor"
```
tonyespy: Since the other two constants end with the actual encoding mechanism (e.g. Json, and Yaml), shouldn't this one too?

Also in general, names which include initialisms (or acronyms) should uses consistent case (e.g. "URL" or "url", not "urL" or "Url"). See [Go Code Review comments](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

brandonforster:
> This is looking pretty good although I did have one comment about the ContentType constant names.
> 
> The only other thing that comes to mind is that issue #32 mentions that MediaType needs to be exposed in ValueDescriptor.

https://github.com/edgexfoundry/go-mod-core-contracts/issues/32#issuecomment-473116895

tonyespy:
> > This is looking pretty good although I did have one comment about the ContentType constant names.
> > The only other thing that comes to mind is that issue #32 mentions that MediaType needs to be exposed in ValueDescriptor.
> 
> [#32 (comment)](https://github.com/edgexfoundry/go-mod-core-contracts/issues/32#issuecomment-473116895)

So here's the deal, having MediaType in the ProfileProperty is required, as that's part of the DeviceResource struct, however if a client of Core Data wants to understand the contents of a specific Reading, the MediaType should also be present in the value descriptor. Without this, a client would be forced to search thru device profiles in order to find the associated DeviceResource ProfileProperty See the following edgex-go issue for more details:

https://github.com/edgexfoundry/edgex-go/issues/1071

PR #37, opened 2019-03-18
Added missing field to ValueDescriptor.

brandonforster:


models/value-descriptor.go
```diff
@@ -35,6 +35,7 @@ type ValueDescriptor struct {
 	UomLabel     string      `json:"uomLabel"`
 	Formatting   string      `json:"formatting"`
 	Labels       []string    `json:"labels"`
+	MediaType    string      `json:"mediaType"`
```
tsconn23: Include `omitempty` here then LGTM

Issue #36, opened 2019-03-18
Add CBOR Conversion to AddEvent


tsconn23:
Pending merge of PR #35.

The [Add() function](https://github.com/edgexfoundry/go-mod-core-contracts/blob/7a2734aa6c016ab61519f629ae24eedac203b52f/clients/coredata/event.go#L133) of event.go within the core-data client should check the incoming Context for a `ContentType` value. If a value is not found, then `ContentTypeJson` should be the default. However if `ContentTypeCbor` is found, then the client should marshal the incoming event as CBOR instead of JSON. This should still allow the client to pass a byte array to the `clients.PostRequest()` method.

The current candidate for accomplishing CBOR marshalling is [ugorji](https://github.com/ugorji/go). This library has no dependencies other than the Go std lib.

brandonforster:
closed via #39

PR #39, opened 2019-03-19
Events can now be marshalled as CBOR.

brandonforster:
Issue #36 

### DIFF
--- a/clients/coredata/event.go
+++ b/clients/coredata/event.go
@@ -131,7 +131,12 @@ func (e *EventRestClient) EventsForDeviceAndValueDescriptor(deviceId string, vd 
 
 // Add event
 func (e *EventRestClient) Add(event *models.Event, ctx context.Context) (string, error) {
-	return clients.PostRequest(e.url, []byte(event.String()), ctx)
+	content := clients.FromContext(clients.ContentType, ctx)
+	if content == clients.ContentTypeCBOR {
+		return clients.PostRequest(e.url, event.CBOR(), ctx)
+	} else {
+		return clients.PostJsonRequest(e.url, event, ctx)
+	}
 }
 
 // Delete event by id
--- a/models/event.go
+++ b/models/event.go
@@ -16,6 +16,8 @@ package models
 
 import (
 	"encoding/json"
+
+	"github.com/ugorji/go/codec"
 )
 
 /*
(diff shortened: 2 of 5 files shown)


clients/coredata/event.go
```diff
@@ -131,7 +131,20 @@ func (e *EventRestClient) EventsForDeviceAndValueDescriptor(deviceId string, vd
...
 func (e *EventRestClient) Add(event *models.Event, ctx context.Context) (string, error) {
-	return clients.PostRequest(e.url, []byte(event.String()), ctx)
+	content := clients.FromContext(clients.ContentType, ctx)
+
+	var data []byte
+	var err error
+	if content == clients.ContentTypeCBOR {
+		data, err = event.MarshalCBOR()
+	} else {
+		data, err = event.MarshalJSON()
```
tsconn23: You don't want to call this directly. It's an implementation method for the Marshaler interface that is called by `json.Marshal`

clients/coredata/event.go
```diff
@@ -131,7 +131,20 @@ func (e *EventRestClient) EventsForDeviceAndValueDescriptor(deviceId string, vd
 
 // Add event
 func (e *EventRestClient) Add(event *models.Event, ctx context.Context) (string, error) {
-	return clients.PostRequest(e.url, []byte(event.String()), ctx)
+	content := clients.FromContext(clients.ContentType, ctx)
+
+	var data []byte
+	var err error
+	if content == clients.ContentTypeCBOR {
+		data, err = event.MarshalCBOR()
```
tsconn23: See comment below on MarshalJSON. See also question on whether MarshalCBOR() is an interface method called by the library when marshaling or encoding. If it is not an interface method, then wrap the encoding in a "private" function within clients/coredata/event.go itself.

models/event.go
```diff
@@ -67,6 +68,19 @@ func (e Event) MarshalJSON() ([]byte, error) {
 	return json.Marshal(test)
 }
 
+func (e Event) MarshalCBOR() ([]byte, error) {
```
tsconn23: Is this a similar function in use to MarshalJSON()? By that I mean does this implement an interface like [json.Marshaler](https://golang.org/pkg/encoding/json/#Marshaler)? If so, it shouldn't be called directly. If not, I wouldn't utilize this approach because it's conflating an existing convention.

Issue #31, opened 2019-03-12
Audit GoDoc Compatible Comments


tsconn23:
Do a module-wide audit of [GoDoc](https://blog.golang.org/godoc-documenting-go-code) conformant comments in order to provide updated online documentation.

PR #43, opened 2019-03-20
Formatting GoDoc content for CommandClient

tsconn23:
#31 

First small PR towards the above issue. In addition to revising GoDoc content I made some other small changes that are motivating me to check this in by itself.
- Renamed `command/command.go` to `command/client.go` to avoid stutter
- Made the `commandRestClient` unexported as no application should access that type directly
- Revised the parameter names on Get/Put methods to be more descriptive

It appears this client isn't used in either the Device SDK or EdgeX-go. Indeed its current methods don't reflect the full compliment of the CoreCommand API. Inclusion of this client going forward is probably still worthwhile as someone with the need could use it directly and/or extend it to cover the whole CoreCommand API.

### DIFF
--- a/clients/command/client.go
+++ b/clients/command/client.go
@@ -21,25 +21,25 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 )
 
-// CommandClient : client to interact with core command
+// The CommandClient interface defines interactions with the EdgeX Core Command microservice.
 type CommandClient interface {
 	Get(id string, cID string, ctx context.Context) (string, error)
 	Put(id string, cID string, body string, ctx context.Context) (string, error)
 }
 
-type CommandRestClient struct {
+type commandRestClient struct {
 	url      string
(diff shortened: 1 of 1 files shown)


clients/command/client.go
```diff
@@ -56,13 +56,13 @@ func (c *CommandRestClient) init(params types.EndpointParams) {
 	}
 }
 
-// Get : issue GET command
-func (cc *CommandRestClient) Get(id string, cID string, ctx context.Context) (string, error) {
-	body, err := clients.GetRequest(cc.url+"/"+id+"/command/"+cID, ctx)
+// Get issues a GET command targeting the specified device, using the specified command
+func (cc *commandRestClient) Get(device string, command string, ctx context.Context) (string, error) {
```
jduranf: As we are using the IDs, maybe it's more clear to use `deviceID` and `commandID` instead of `device` and `command`.

tsconn23: Done

Issue #48, opened 2019-03-23
MediaType should be part of PropertyValue not ProfileProperty


tonyespy:
This ones on me... as I suggested the wrong struct in my review comment for PR #35 where I said:

> So here's the deal, having MediaType in the ProfileProperty is required, as that's part of the DeviceResource struct, however if a client of Core Data wants to understand the contents of a specific Reading, the MediaType should also be present in the value descriptor. 

ProfileProperty is a struct which contains PropertyValue and Units.  The former describes the type of the device resource and latter describes the units used to label it. MediaType should actually be part of PropertyValue, as it's another attribute which describes the value of the device resource. The device service SDKs use the PropertyValue to create a ValueDescriptor instance.

brandonforster:
closed via #52

PR #52, opened 2019-03-25
Moved MediaType field from ProfileProperty to PropertyValue

brandonforster:
Issue #48 


brandonforster:
Also included are some `go fmt` changes that it looks like I missed from my other CBOR PR. They can be removed if requested but I think it's worth picking up the formatting changes here.

PR #57, opened 2019-03-29
Renamed the BaseObject type to Timestamps to better reflect its intent

AnthonyMBonafide:
Updated all references to BaseObject to now use Timestamps

### DIFF
--- a/clients/logger/logger.go
+++ b/clients/logger/logger.go
@@ -44,12 +44,7 @@ const (
 	ErrorLog = "ERROR"
 )
 
-var logLevels = []string{
-	TraceLog,
-	DebugLog,
-	InfoLog,
-	WarnLog,
-	ErrorLog}
+var logLevels = LogLevels()
 
 // LoggingClient defines the interface for logging operations.
 type LoggingClient interface {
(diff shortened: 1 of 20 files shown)


AnthonyMBonafide:
I also made a change regarding the logLevels array and its accessibility. When I updated edgex-go to leverage these changes there was a compilation issue due to the logLevels having been updated to not be exported as a global variable. I created an exported function in the LogLevels in the logger package so that the log level information can be safely be shared across repos. I will have changes in the edgex-go repo which will update the to using a version of go-mod-core-contracts which contains these changes, once they are merged.

AnthonyMBonafide:
Please hold off on any reviews for this pull request. I will need to verify the side effects in the `device-sdk-go` project before we can consider merging.

SteveOss:
Apologies if I've missed a discussion on this, but is it not possible that we may want to store other common date in baseobject, other than timestamps, so the existing naming is actually better?

AnthonyMBonafide:
@SteveOss, I think the name `BaseObject` loses its value within GoLang as we are not using inheritance, but rather embedding the type. I believe that is the main reason for the name change. As for the encapsulated data in `BaseObject`/`Timestamps`, did you have any specific data that you think would fit into this type? I am wondering if there is data that needs to be captured in this type and/or if we add more date time attributes will they belong to the `BaseObject`/`Timestamps` type or will be grouping them together in a new type.

SteveOss:
@tsconn23 Was the intent to store anything other than timestamps ? If not then the name change makes sense.

AnthonyMBonafide:
I have to re-verify the changes in `edgex-go` with blackbox-tests, as I did not do it correctly initially.

brandonforster:
The merge conflicts need addressing but it LGTM

PR #58, opened 2019-04-01
Updated GoDoc content on first 17 models

tsconn23:
#31 continued

Updating the GoDoc content of the models in the contracts module in small doses. I can only take so much.

### DIFF
--- a/models/callbackalert.go
+++ b/models/callbackalert.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 )
 
+// CallbackAlert indicates an action to take when a callback fires.
 type CallbackAlert struct {
 	ActionType ActionType `json:"type"`
 	Id         string     `json:"id"`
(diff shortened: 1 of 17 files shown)


models/devicereport.go
```diff
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 )
 
+// DeviceReport has been deprecated and needs to be removed.
```
brandonforster: It looks like there is some style associated with deprecation, namely, having a line starting with `"Deprecated: "` and explaining what to use instead or what the user should be trying to accomplish instead of using this.

[This blog post has some links to Go standard library docs making use of the convention with discussion on the `golang` repo](https://rakyll.org/deprecated/)

[Here's the official wiki on the matter.](https://github.com/golang/go/wiki/Deprecated)

tsconn23: Reformatted content using guidelines above.

models/callbackalert.go
```diff
@@ -39,6 +40,9 @@ func (ca CallbackAlert) MarshalJSON() ([]byte, error) {
 	return json.Marshal(test)
 }
 
+/*
+ * String function for representing a CallbackAlert
```
brandonforster: What was the rationale for using a block comment here?

tsconn23: On all types String() has a block comment so I made it the same here rather than changing them all

Issue #27, opened 2019-03-10
Consider removal of DeviceProfile from Device struct


tonyespy:
I recently reported an [issue](https://github.com/edgexfoundry/device-sdk-go/issues/202) against device-sdk-go that concerned dumping entire objects in non-debug log messages.

The offending log message was actually including the entire state of a Device object, which is enormous! This got me to thinking whether it's actually a good idea for us to be embedding an entire DeviceProfile in every device as opposed to DeviceProfile.Name? The device SDKs have a cache of device profiles, and to me, it would be much more efficient to always retrieve them from the cache instead of embedding them in every single device.

It also appears that the device-sdk-go [callback handler](https://github.com/edgexfoundry/device-sdk-go/blob/master/internal/handler/callback.go#L105) doesn't properly update every device in its device cache when a Profile is updated. It's also not clear who's responsible to update value descriptors if/when a DeviceProfile change occurs. It doesn't look like device-sdk-go currently handles this case.

Note - this is related to issue #40 which proposes removing DeviceService from Device.

difince:
If you take the decision to make the Device struct lighter (and I think it is doable from DS point of view, at least I took a look of device.DeviceProfile struct) can I work on it? I think it is good issue for a newcomer. To resolve this issue a lot of places will need to be touched - giving a good overview and understanding of the project implementation.

tsconn23:
@difince You are certainly welcome to give it a shot, however the teams are very focused on specific tasks leading to the Edinburgh deliverable right now. That means I can't promise you when a potential change might get pulled in. If you do decide to investigate this further, remember that this contract module is utilized by edgex-go, device-sdk-go and app-functions-sdk-go. In particular with those first two repos, I expect the above change will cause side effects that will also need to be addressed. As part of your work, if you don't mind, please try to document those side effects here as part of this issue because if we were to merge your change, we would then be committed to doing the integration work in the other repos.

For more context you're welcome to join the weekly [Device Service](https://wiki.edgexfoundry.org/display/FA/Device+and+Device+SDK+Working+Group) and [Core Working Group](https://wiki.edgexfoundry.org/display/FA/Core+Working+Group) calls.

jduranf:
I completely agree with the idea, just some comments:
1. Will be fine if the equivalent change also it's done with Provision Watchers
[...]

PR #63, opened 2019-04-08
Replace device.DeviceProfile with device.ProfileName

difince:
Remove the embedded DeviceProfile struct from the Device.
The Device struc will keep just the ProfileName as a reference.
The Device SDK have a cache of Device Profiles, so they will be retrieved from the cache.
All other service will make a call to metadata in order to get a Device Profile.

It decreases the Device SDK cache size, and descrease the Debug and error log messages.

Resolves/Fixes edgexfoundry/go-mod-core-contracts#27

Signed-off-by: Diana Atanasova <dianaa@vmware.com>
Signed-off-by: difince <dianaa@vmware.com>


models/device.go
```diff
@@ -31,7 +31,7 @@ type Device struct {
 	Labels         []string                      `json:"labels"`         // Other labels applied to the device to help with searching
 	Location       interface{}                   `json:"location"`       // Device service specific location (interface{} is an empty interface so it can be anything)
 	Service        DeviceService                 `json:"service"`        // Associated Device Service - One per device
-	Profile        DeviceProfile                 `json:"profile"`        // Associated Device Profile - Describes the device
+	ProfileName    string                        `json:"profileName""`	 // Associated Device Profile Name
```
tsconn23: Syntax error on this line with trailing double quotes

difince: Hi @tsconn23, thanks for your review. Unfortunately, most probably for today Core meeting I will be late. Hope, I will join for the second half.

Issue #67, opened 2019-04-15
Create AdminState and OperatingState Request Types


tsconn23:
Changes are required to facilitate the functionality described in the following edgex-go issue
https://github.com/edgexfoundry/edgex-go/issues/1244

This issue will give us the opportunity to introduce functional "request types" into our contracts -- that is, request models that are specific to a particular API endpoint rather than general models which are utilized in various ways for various endpoints. Our current contract models were originally state models within edgex-go. They have been refactored out into their own module now, which is good, but we need to go further so that requests become expressions of intent and support encapsulated validation (per https://github.com/edgexfoundry/go-mod-core-contracts/issues/66).

An example of the above, albeit for a different use case, can be found in the following feature branch.
https://github.com/tsconn23/go-mod-core-contracts/tree/marshal_bug/requests/provision_watcher

Utilizing the package naming [best practices](https://blog.golang.org/package-names), I would suggest creating a `requests` package at the same level as `models` and `clients`. Then underneath that, create a package structure like the following

`states`
- `admin`
- `operating`

The `admin` and `operating` packages above should each contain an `update.go` wherein the respective `UpdateRequest` is defined. The request's sole member would be either `models/OperatingState` or `models/AdminState` so that JSON marshaling will create a request that looks like this:
`{"adminState":"unlocked"}`
`{"operatingState":"disabled"}`

There should also be tests, as in the example branch, but we'll need to see how the timing works out between this issue and #66 which is implementing a new comprehensive model validation scheme. In effect, the work of #66 (which you can see in the feature branch linked in the comments there) will allow these new request types to call Validate() on their respective `AdminState` or `OperatingState` members. If the work on this effort completes first, we may just write the tests as part of #66.

brandonforster:
closed via #69 (nice)

brandonforster:
Closed again via #73

PR #69, opened 2019-04-17
Created AdminState and OperatingState request types.

brandonforster:
Issue #67


models/interval.go
```diff
@@ -37,20 +37,14 @@ func (i Interval) MarshalJSON() ([]byte, error) {
...
 		End       *string `json:"end,omitempty"`       // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
 		Frequency *string `json:"frequency,omitempty"` // how frequently should the event occur
 		Cron      *string `json:"cron,omitempty"`      // cron styled regular expression indicating how often the action under schedule should occur.  Use either runOnce, frequency or cron and not all.
 		RunOnce   bool    `json:"runOnce,omitempty"`   // boolean indicating that this interval runs one time - at the time indicated by the start
 	}{
-		Created:  i.Created,
-		Modified: i.Modified,
-		Origin:   i.Origin,
-		RunOnce:  i.RunOnce,
+		RunOnce: i.RunOnce,
```
tsconn23: There's a discrepancy here. The Timestamps type is a struct that contains Created, Modified and Origin properties. These have been removed as separate properties from the marshaling struct, but not from the Interval type itself. Also, you would need to set the Timestamps struct so the values are retained
```
test := struct {
    Timestamps
    ...
}{
    Timestamps: i.Timestamps
    ...
}
```

models/interval.go
```diff
@@ -20,37 +20,30 @@ import (
...
-	Created   int64  `json:"created"`
-	Modified  int64  `json:"modified"`
-	Origin    int64  `json:"origin"`    // not currently used reserved for interval origin source
-	Name      string `json:"name"`      // non-database identifier for a shcedule (*must be quitue)
-	Start     string `json:"start"`     // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
-	End       string `json:"end"`       // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
-	Frequency string `json:"frequency"` // how frequently should the event occur according ISO 8601
-	Cron      string `json:"cron"`      // cron styled regular expression indicating how often the action under interval should occur.  Use either runOnce, frequency or cron and not all.
-	RunOnce   bool   `json:"runOnce"`   // boolean indicating that this interval runs one time - at the time indicated by the start
+	Timestamps Timestamps `json:"timestamps"`
```
tsconn23: This line doesn't require a JSON tag

models/interval.go
```diff
@@ -20,37 +20,30 @@ import (
...
-		Created   int64   `json:"created,omitempty"`
-		Modified  int64   `json:"modified,omitempty"`
-		Origin    int64   `json:"origin,omitempty"`    // not currently used reserved for interval origin source
-		Name      *string `json:"name,omitempty"`      // non-database identifier for a shcedule (*must be quitue)
-		Start     *string `json:"start,omitempty"`     // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
-		End       *string `json:"end,omitempty"`       // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
-		Frequency *string `json:"frequency,omitempty"` // how frequently should the event occur
-		Cron      *string `json:"cron,omitempty"`      // cron styled regular expression indicating how often the action under schedule should occur.  Use either runOnce, frequency or cron and not all.
-		RunOnce   bool    `json:"runOnce,omitempty"`   // boolean indicating that this interval runs one time - at the time indicated by the start
+		Timestamps Timestamps `json:"timestamps,omitempty"`
```
tsconn23: For this line, you'll want to use `json:",inline"` without the `omitempty`. This will inline the properties of the Timestamps type into the Interval when it is marshalled.

PR #77, opened 2019-04-25
Fixed Event Struct Tags

AnthonyMBonafide:
This commit fixes issue 76

Updated the struct tags on the Event type. Previously, the id portion of
the struct tag was not specified for the "codec" type, which was causing
issues when encoding and decoding into CBOR using the codec library.

Added "codec" struct tags to the Reading type

Added tests for both types which validates CBOR encoding and decoding

### DIFF
--- a/models/event.go
+++ b/models/event.go
@@ -23,13 +23,13 @@ import (
 
 // Event represents a single measurable event read from a device
 type Event struct {
-	ID          string    `json:"id" codec:"omitempty"`       // ID uniquely identifies an event, for example a UUID
-	Pushed      int64     `json:"pushed" codec:"omitempty"`   // Pushed is a timestamp indicating when the event was exported. If unexported, the value is zero.
-	Device      string    `json:"device" codec:"omitempty"`   // Device identifies the source of the event, can be a device name or id. Usually the device name.
-	Created     int64     `json:"created" codec:"omitempty"`  // Created is a timestamp indicating when the event was created.
-	Modified    int64     `json:"modified" codec:"omitempty"` // Modified is a timestamp indicating when the event was last modified.
-	Origin      int64     `json:"origin" codec:"omitempty"`   // Origin is a timestamp that can communicate the time of the original reading, prior to event creation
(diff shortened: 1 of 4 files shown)


models/event.go
```diff
@@ -23,13 +23,13 @@ import (
...
-	Created     int64     `json:"created" codec:"omitempty"`  // Created is a timestamp indicating when the event was created.
-	Modified    int64     `json:"modified" codec:"omitempty"` // Modified is a timestamp indicating when the event was last modified.
-	Origin      int64     `json:"origin" codec:"omitempty"`   // Origin is a timestamp that can communicate the time of the original reading, prior to event creation
-	Readings    []Reading `json:"readings" codec:"omitempty"` // Readings will contain zero to many entries for the associated readings of a given event.
+	ID          string    `json:"id" codec:"id,omitempty"`             // ID uniquely identifies an event, for example a UUID
+	Pushed      int64     `json:"pushed" codec:"pushed,omitempty"`     // Pushed is a timestamp indicating when the event was exported. If unexported, the value is zero.
+	Device      string    `json:"device" codec:"device,omitempty"`     // Device identifies the source of the event, can be a device name or id. Usually the device name.
+	Created     int64     `json:"created" codec:",created,omitempty"`  // Created is a timestamp indicating when the event was created.
```
tsconn23: Is that leading comma in the codec tag supposed to be there?

models/event_test.go
```diff
@@ -15,6 +15,7 @@
 package models
 
 import (
+	"github.com/ugorji/go/codec"
```
tsconn23: Move this down to the bottom of the imports, after the std lib packages

models/reading_test.go
```diff
@@ -17,15 +17,17 @@ package models
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/ugorji/go/codec"
```
tsconn23: Move this down to the bottom of the imports, after the std lib packages

models/event.go
AnthonyMBonafide: No, that is a mistake. Great catch.

Issue #62, opened 2019-04-08
Proposal -- Core-data.AddEvent() to Receive Byte Array


tsconn23:
1.) Add new method to EventClient interface
     EncodeEvent(e models.Event) ([]byte, error)
2.) Modify EventClient.Add([]byte, Context) (string, error)

Based on discussion around this flow https://github.com/edgexfoundry/device-sdk-go/issues/233

andresrinivasan:
I read through [edgexfoundry/device-sdk-go#233](https://github.com/edgexfoundry/device-sdk-go/issues/233) and I'm not following how this proposal addresses the issue. Can you elaborate a bit more?

tonyespy:
If I'm interpreting @tsconn23's description, a client wishing to push an Event to Core Data would call the EventClient.EncodeEvent() method to delegate the encoding to the EventClient, then instead if passing an Event to EventClient.Add, it would pass an already encoded Event as a byte array.

tsconn23:
@tonyespy @tobiasmo1 I think I'm going to add another method to the EventClient for the Add. In testing, it's extremely convenient to be able to issue an AddEvent via POST containing JSON. That path would continue to work as it does today, and then I would provide another method `EventClient.AddBytes([]byte, Context) (string, error)`.

Let me know if you have objections.

PR #84, opened 2019-05-01
Added AddBytes & MarshalEvent methods to EventClient

tsconn23:
#62 

The Device Services group asked that the EventClient take on the centralized responsibility for handling encoding of events. To that end, marshaling has been exposed to callers via the MarshalEvent method. After the caller uses this method to encode the event, they can add the event directly as a byte array via the new EventClient.AddBytes method.


clients/coredata/event.go
```diff
@@ -149,6 +154,10 @@ func (e *eventRestClient) Add(event *models.Event, ctx context.Context) (string,
 	}
 }
 
+func (e *eventRestClient) AddBytes(event []byte, ctx context.Context) (string, error) {
```
tobiasmo1: Note that responsibility is on the caller to specify context (esp content type) for AddBytes method.

Issue #40, opened 2019-03-20
Consider removal of DeviceService from Device struct


tonyespy:
I recently reported an issue against device-sdk-go that concerned dumping entire objects in non-debug log messages.

The offending log message was actually including the entire state of a Device object, which is enormous! This got me to thinking whether it's actually a good idea for us to be embedding an entire DeviceService in every device as opposed to DeviceService.Name (or ID)? 
Note, this is related to issue #27 which proposes removing DeviceProfile from Device.

michaelestrin:
Closed.  Expect this to be addressed by the [Geneva v2 API](https://github.com/tsconn23/edgex-geneva-api) changes.

PR #86, opened 2019-05-02
In some contexts a device service struct may only have its name populated

iain-anderson:
In particular, when it is marshaled as part of a Device.
This situation will go away if #40 is implemented: we could then perhaps return to full validation

### DIFF
--- a/models/deviceservice.go
+++ b/models/deviceservice.go
@@ -16,6 +16,7 @@ package models
 
 import (
 	"encoding/json"
+	"errors"
 )
 
 // DeviceService represents a service that is responsible for proxying connectivity between a set of devices and the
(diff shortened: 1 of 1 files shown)


iain-anderson:
created #87

tsconn23:
I'm going to close this PR since we have decided on another way of solving this issue.
https://github.com/edgexfoundry/device-sdk-go/pull/271
https://github.com/edgexfoundry/edgex-go/issues/1311

cloudxxx8:
@tsconn23 This PR is about validating Device Service struct validation, so I am not sure this is related to the Device name in Readings
edgexfoundry/device-sdk-go#271
edgexfoundry/edgex-go#1311

Issue #85, opened 2019-05-02
Return Custom ErrContractInvalid Error Type When Validation Fails


tsconn23:
When the contract models fail to validate their internal state, they should return a specific error type to allow for more specific error handling within the API controller functions. For example, when performing unmarshaling within a controller, a type check on the returned error will allow us to be more explicit about whether we should return a `400 Bad Request` in the case of a model's invalid state versus some unexpected condition. The latter would result in a `500`. response.

PR #88, opened 2019-05-05
Added ErrContractInvalid Type. Return from Validate()

tsconn23:
#85

- Created ErrContractInvalid type, implementing Go stdlib `error`
  interface.
- The purpose of this type is to allow a caller to differentiate
  explicitly between a problem with validation versus an unexpected error
- For example, an API router could type check a returned error and if
  it's of type ErrContractInvalid, return a 400. Otherwise return 500.
- For tests oriented around validation testing, created a new
  checkValidationError() function for standard handling. Integrated
  into existing tests.
- Corrected test issue whereby a type instance such as TestAddressable
  that is used in multiple tests could cause tests to fail if used
  directly. This is because the private isValidated member would be set
  to true by the first test, causing validation to not be run by
  subsequent tests which would produce erroneous results. The solution
  is simply to copy the TestAddressable into a local variable (called
  "valid" where the change was made) and then use that within the test.

### DIFF
--- a/models/addressable.go
+++ b/models/addressable.go
@@ -17,7 +17,6 @@ package models
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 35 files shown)


models/addressable.go
```diff
@@ -184,9 +184,10 @@ func (a *Addressable) UnmarshalJSON(data []byte) error {
 
 // Validate satisfies the Validator interface
 func (a Addressable) Validate() (bool, error) {
+	fmt.Println(fmt.Sprintf("Addressable.Validate isValidated=%v", a.isValidated))
```
jduranf: If I'm not wrong, this line should be removed

tsconn23: Oh, good catch! Yes, this was part of my debugging when I was trying to understand why tests were failing. See fourth bullet in PR description.

Issue #96, opened 2019-05-08
Remove Validation of Reading.Device Property


tsconn23:
Source issue https://github.com/edgexfoundry/device-sdk-go/issues/269

In order to accomplish this, the following validation requirement that a reading must have a Device name assigned has to be relaxed
https://github.com/edgexfoundry/go-mod-core-contracts/blob/b02b9d963a25310697254f12fbd44cff18e10242/models/reading.go#L126

Following from the above, there's already a PR in edgex-go which will assign the Device name from the Event to each reading before persisting to the database.
https://github.com/edgexfoundry/edgex-go/pull/1324/

Once PR #1324 is merged and this issue is addressed, then device-sdk-go https://github.com/edgexfoundry/device-sdk-go/pull/271 should be ready for integration.

brandonforster:
closed via #97

PR #97, opened 2019-05-08
Removing Required Device Name from Reading

tsconn23:
#96 

### DIFF
--- a/models/reading_test.go
+++ b/models/reading_test.go
@@ -95,7 +95,7 @@ func TestReadingValidation(t *testing.T) {
 		expectError bool
 	}{
 		{"valid reading", valid, false},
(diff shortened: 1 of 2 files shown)


models/reading_test.go
```diff
@@ -95,7 +95,7 @@ func TestReadingValidation(t *testing.T) {
 		expectError bool
 	}{
 		{"valid reading", valid, false},
-		{"invalid device", Reading{Name: "test", Value: "0"}, true},
+		{"invalid device", Reading{Name: "test", Value: "0"}, false},
```
cloudxxx8: It's better to rename this test to "empty device" instead of "invalid device"

PR #100, opened 2019-05-20
Fix Interval's Validate() method to check timestamp

jdharms:
Fix #1308

Interval's Validate() method was attempting to parse an
integer out of a timestamp.  This change has it parsing
a time using the time package's Parse() function.


jdharms:
Updated unit tests to reflect changes; moved constant out of function and into file's constant declarations.

PR #110, opened 2019-06-14
Add new func to CommandClient interface

difince:
Add CommandsForDevice func to CommandClient Interface
Command service uses it when try to get device commands. 

Signed-off-by: difince <dianaa@vmware.com>


clients/metadata/command.go
```diff
@@ -35,6 +35,8 @@ type CommandClient interface {
 	Commands(ctx context.Context) ([]models.Command, error)
 	// CommandsForName lists all the commands for the specified name
 	CommandsForName(name string, ctx context.Context) ([]models.Command, error)
+	// CommandForDevice obtains all commands of a device with specified ID
+	CommandsForDevice(did string, ctx context.Context) ([]models.Command, error)
```
tsconn23: - Let's call this `CommandsForDeviceId` (the function above it should ideally be called `CommandsForDeviceName` but don't modify as that would be a breaking change). 
- Because the function name will then be explicit, change the first param from `did` to `id`
- Also be sure in your GoDoc comment you write `Commands` plural instead of `Command`

difince: Done

PR #123, opened 2019-07-27
Fix ValueDescriptorClient Usage Method

### LINKED ISSUE #122: ValueDescriptorClient does not include query parameters for the Usage API
Update the logic which builds the URI for the usage API so that it properly sends the ValueDescriptor names.

Here is the problematic method:
https://github.com/edgexfoundry/go-mod-core-contracts/blob/e063b4f6f09902706a34e8a2f9899bba217a1d2c/clients/coredata/value_descriptor.go#L135

This is not a major issue as it is a new API which was added for the ongoing work to migrate the SDK ValueDescriptor logic to core-metadata.

AnthonyMBonafide:
Fix #122

Fixed issue with the generation of URL not including the names query
parameters

### DIFF
--- a/clients/coredata/value_descriptor.go
+++ b/clients/coredata/value_descriptor.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/url"
+	"strings"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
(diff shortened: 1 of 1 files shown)


AnthonyMBonafide:
I discovered this while manual verifying my current work for https://github.com/edgexfoundry/edgex-go/issues/1471 . This has been integrated with `edgex-go` locally, manually tested, and verifified

PR #126, opened 2019-08-09
Update ProvisionWatcher model as per #111

### LINKED ISSUE #111: ProvisionWatcher: replace OperatingState field, add blocking criteria
The purpose of the OperatingState field was that devices created as a result of ProvisionWatcher matches would initially be either enabled or disabled.
However:
* The Administrative State is more applicable to controlling availability for policy reasons.
* We would like to allow for blacklisting using the ProvisionWatcher mechanism.

So:
1) Define a new enumeration "OnMatch" which defines the action to take when a PW is matched. Available values to be
* IGNORE - matching devices are not added
* LOCK - matching devices to be added in the Locked AdminState
* UNLOCK - matching devices to be added in the Unlocked AdminState
2) Replace the OperatingState field with one of this type.

Device services should then, on discovery of a new device, proceed according to the most restrictive PW that matches on that device.

**EDIT: Finalized ProvisionWatcher design, replacing the above, is here: [https://github.com/edgexfoundry/go-mod-core-contracts/issues/111#issuecomment-517362872](url)**

iain-anderson:
Fix #111

### DIFF
--- a/models/provisionwatcher.go
+++ b/models/provisionwatcher.go
@@ -20,31 +20,34 @@ import (
 
 type ProvisionWatcher struct {
 	Timestamps
-	Id             string            `json:"id"`
-	Name           string            `json:"name"`           // unique name and identifier of the addressable
-	Identifiers    map[string]string `json:"identifiers"`    // set of key value pairs that identify type of of address (MAC, HTTP,...) and address to watch for (00-05-1B-A1-99-99, 10.0.0.1,...)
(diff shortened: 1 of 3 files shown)


tsconn23:
The code on this looks fine, and that's an inventive solution w/r/t the `validate` struct tag. Let's discuss this on Monday's Device Svcs call before I merge. The changes are backward compatible but if merged now will be included in the Fuji release (projected to be an LTS) where they won't really have any significance. That's not necessarily a blocker since the whole ProvisionWatcher is incomplete right now. I just think to think on it over the weekend.

Issue #130, opened 2019-09-03
Transmission Add Unmarshalling Validation


nkstanchev:
The Transmission struct should be validated when being unmarshalled as stated in edgex-go's api/support-notifications.raml file

PR #131, opened 2019-09-03
Transmission Add Unmarshalling Validation

nkstanchev:
Addresses #130.

Signed-off-by: Nikolay Stanchev <nstanchev@vmware.com>

### DIFF
--- a/models/transmission.go
+++ b/models/transmission.go
@@ -28,6 +28,7 @@ type Transmission struct {
 	Status       TransmissionStatus   `json:"status,omitempty"`
 	ResendCount  int                  `json:"resendcount"`
 	Records      []TransmissionRecord `json:"records,omitempty"`
+	isValidated  bool
 }
 
 // Custom marshaling to make empty strings null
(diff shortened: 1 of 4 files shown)


models/transmission.go
```diff
@@ -59,6 +60,74 @@ func (t Transmission) MarshalJSON() ([]byte, error) {
...
+	if a.ID != nil {
+		t.ID = *a.ID
+	}
+	t.Notification = a.Notification
+
+	if a.Receiver != nil {
+		t.Receiver = *a.Receiver
+	}
+	t.Channel = a.Channel
+	t.Status = a.Status
```
tsconn23: If you're going to go down this validation path, then the value of Status should also be checked to ensure it's in the range of acceptable values. We have an example of that if you look into how OperatingState and AdminState are validated on the Device
https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/device.go

ernestojeda:
recheck

models/transmission.go
nkstanchev: I think Notification, Channel and Channel's ChannelType would also need to have a Validate method and unit tests. Didn't fully grasp the volume of the required work but I can't get properly tested support services without this. Should these be in a different PR?

tsconn23: `Validate the NotificationStatus` as part of this PR and then create an issue for the follow-up validation of the dependent types. Once the issue is created, let me know and I can assign to you

models/transmission_status.go
```diff
@@ -39,13 +40,19 @@ func (as *TransmissionStatus) UnmarshalJSON(data []byte) error {
...
 	return nil
 }
+
+func (as TransmissionStatus) Validate() (bool, error) {
+	_, found := map[string]TransmissionStatus{"FAILED": Failed, "SENT": Sent, "ACKNOWLEDGED": Acknowledged, "TRXESCALATED": Trxescalated}[string(as)]
+	if !found {
+		return false, NewErrContractInvalid(fmt.Sprintf("invalid Transmission Status %q", as))
+	}
+	return true, nil
+}
```
tsconn23: Needs unit test

tsconn23: Also transmission.go::UnmarshalJSON() needs to be covered by unit test. That's where the large drop in coverage reported by codecov.io is coming from.

lranjbar:
recheck

tsconn23:
@nkstanchev Needs go fmt
```
17:06:04  CGO_ENABLED=0 GO111MODULE=on go vet ./...
17:06:26  gofmt -l .
17:06:26  models/transmission_test.go
17:06:26  [ "`gofmt -l .`" = "" ]
17:06:26  make: *** [Makefile:9: test] Error 1
```

tsconn23:
recheck

PR #132, opened 2019-09-03
Implement the design for when an External Entity Requests SMA to Set …

### LINKED ISSUE #128: External Entity Requests SMA to Set Configuration.
- Implement the design: External Entity Requests SMA to Set Configuration .
- This is a direct follow-up to: [Issue #1374](https://github.com/edgexfoundry/edgex-go/issues/1374)
- This replace [Issue #1430](https://github.com/edgexfoundry/edgex-go/issues/1430).

akramtexas:
Fix #128 

- This PR has the change-set which implement the design for when an external entity requests the SMA to set configuration . 
- This work lays the foundation for other work in progress which implements the design for when the SMA itself goes about servicing the request of setting the configuration (writable) without restart  (covered by [Issue-1432](https://github.com/edgexfoundry/edgex-go/issues/1432)).


akramtexas:
I have defined a specific type that is returned in response to a set config request. Incorporating it is a `WIP`.

akramtexas:
recheck

requests/configuration/set.go
```diff
@@ -0,0 +1,84 @@
...
+		Key   *string `json:"key"`
+		Value *string `json:"value"`
+	}{}
+
+	//Verify that incoming string will unmarshal successfully
+	if err = json.Unmarshal(data, &test); err != nil {
+		return err
+	}
+
+	sc.isValidated, err = sc.Validate()
```
tsconn23: `Validate()` should be called after the properties on `sc` are set. Otherwise the instance is effectively empty.

Example
https://github.com/edgexfoundry/go-mod-core-contracts/blob/4a4a40fd862f5e9270a97718ee09393586d09a2a/models/device.go#L127

Also, note how `err` is returned in that example. For the code here, the `err` being returned from the call to `Validate()` is lost.

requests/configuration/set_test.go
```diff
@@ -0,0 +1,50 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package configuration
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+	"testing"
```
tsconn23: Organize imports

requests/configuration/set.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+type Key string
+
+/*
+ * SetConfigRequest is for SMA processing of incoming (PUT) requests to Set Config.
+ *
+ * SetConfigRequest and SetConfigResponse struct each follows
```
tsconn23: Omit this line and just put the line above as a comment preceded by `//` immediately above the type declaration. `SetConfigResponse` isn't defined here.

responses/configuration/set.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"encoding/json"
+	"reflect"
+	"strings"
+)
+
+/*
+ * SetConfigResponse is for SMA to use when responding to an incoming request to PUT (i.e. UPDATE) a resource.
+ *
+ */
```
tsconn23: Define using `//` immediately above the type definition (no blank line)

responses/configuration/set.go
```diff
@@ -0,0 +1,67 @@
...
+
+type SetConfigResponse struct {
+	Code           string   `json:"code" yaml:"code,omitempty"`
+	Description    string   `json:"description" yaml:"description,omitempty"`
+	ExpectedValues []string `json:"expectedValues" yaml:"expectedValues,omitempty"`
```
tsconn23: For each property can you add a comment to the right briefly describing each one for GoDoc purposes?

requests/configuration/set.go
```diff
@@ -0,0 +1,84 @@
...
+ */
+type SetConfigRequest struct {
+	Key         string `bson:"key" json:"key,omitempty"`
+	Value       string `bson:"value" json:"value,omitempty"`
+	isValidated bool   // internal member used for validation check
```
tsconn23: For each property can you add a comment to the right briefly describing each one for GoDoc purposes?

responses/configuration/set.go
```diff
@@ -0,0 +1,67 @@
...
+ */
+func (a SetConfigResponse) String() string {
+	out, err := json.Marshal(a)
+	if err != nil {
+		return err.Error()
+	}
+	return string(out)
+}
+
+func (r SetConfigResponse) Equals(r2 SetConfigResponse) bool {
```
tsconn23: What is this function for? I don't see it used in the PR.

responses/configuration/set.go
```diff
@@ -0,0 +1,67 @@
...
+	if r.Code != "" {
+		test.Code = &r.Code
+	}
+	if r.Description != "" {
+		test.Description = &r.Description
+	}
+
+	return json.Marshal(test)
+}
+
```
tsconn23: There isn't an `UnmarshalJSON` function in this type. Is that intentional? Could be useful to an application receiving the response.

requests/configuration/set.go
akramtexas: Excellent feedback here and in the remainder of the PR as well, @tsconn23 - Thanks, I'm on it.

PR #149, opened 2019-09-19
 Expand CommandClient interface for Put/Get by Names 

jbonafide623:
Fix #141 

Expand CommandClient interface for Put/Get by Names

- Add GetDeviceCommandByNames method to Command client

- Add PutDeviceCommandByNames method to Command client

- Add unit tests for the client

Signed-off-by: Jason Bonafide <jbonafide623@gmail.com>

PR #155, opened 2019-09-24
Add "/name/" after base api route for command client

### LINKED ISSUE #154: URLs in Get() and Put() functions in clients/command/client.go are not constructed correctly
There should be "/name/" after cc.url as in GetDeviceCommandByNames() and PutDeviceCommandByNames()
@cloudxxx8

davidmsh:
Fix #154

Signed-off-by: David Chang <david@iotechsys.com>

### DIFF
--- a/clients/command/client.go
+++ b/clients/command/client.go
@@ -69,12 +69,12 @@ func (c *commandRestClient) init(params types.EndpointParams) {
 }
 
 func (cc *commandRestClient) Get(deviceId string, commandId string, ctx context.Context) (string, error) {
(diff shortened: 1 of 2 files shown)


tsconn23:
@davidmsh Why are we putting a `name` parameter in front of methods used to interact with  commands by ID? We have other methods in the CoreCommand client that utilize the `/name/` route as of PR https://github.com/edgexfoundry/go-mod-core-contracts/pull/149/files

davidmsh:
@tsconn23 yes you are right. That's because I notice thins in an old branch which does not have the two other functions. I will close this PR

PR #156, opened 2019-09-25
Remove the field ExpectedValues (from SetConfigResponse) as it is redundant.

### LINKED ISSUE #152: Remove the field ExpectedValues (from SetConfigResponse) as it is redundant.
- With reference to the following struct (defined in `go-mod-core-contracts/responses/configuration/set.go`), remove the field ExpectedValues as it is redundant:
```
//SetConfigResponse is for SMA to use when responding to an incoming request to PUT (i.e. UPDATE) a resource.
type SetConfigResponse struct {
	Code           string   `json:"code" yaml:"code,omitempty"`                     //HTTP code to return (e.g. 200, 500, etc.)
	Description    string   `json:"description" yaml:"description,omitempty"`       //info allied with preceding HTTP code
	ExpectedValues []string `json:"expectedValues" yaml:"expectedValues,omitempty"` //expected values to return
	isValidated    bool     //internal member used for validation check
}
```

akramtexas:
Fix #152 
- This is a follow up to #128 (External Entity Requests SMA to Set Configuration).

### DIFF
--- a/responses/configuration/set.go
+++ b/responses/configuration/set.go
@@ -8,27 +8,21 @@ import (
 
 //SetConfigResponse is for SMA to use when responding to an incoming request to PUT (i.e. UPDATE) a resource.
 type SetConfigResponse struct {
-	Code           string   `json:"code" yaml:"code,omitempty"`                     //HTTP code to return (e.g. 200, 500, etc.)
-	Description    string   `json:"description" yaml:"description,omitempty"`       //info allied with preceding HTTP code
-	ExpectedValues []string `json:"expectedValues" yaml:"expectedValues,omitempty"` //expected values to return
-	isValidated    bool     //internal member used for validation check
+	Success     bool   `json:"success"`               // Success or failure of request itself.
+	Description string `json:"description,omitempty"` // Info allied with the preceding result.
+	isValidated bool   // internal member used for validation check
 }
 
 // Custom marshalling to make empty strings null
 func (sc SetConfigResponse) MarshalJSON() ([]byte, error) {
 	test := struct {
-		Code           *string  `json:"code,omitempty"`
-		Description    *string  `json:"description,omitempty"`
-		ExpectedValues []string `json:"expectedValues,omitempty"`
-		isValidated    bool
-	}{
-		ExpectedValues: sc.ExpectedValues,
-	}
+		Success     bool    `json:"success"`
+		Description *string `json:"description,omitempty"`
+		isValidated bool
+	}{}
 
-	// Empty strings are null
-	if sc.Code != "" {
-		test.Code = &sc.Code
-	}
+	test.Success = sc.Success
+
 	if sc.Description != "" {
 		test.Description = &sc.Description
 	}
@@ -49,9 +43,8 @@ func (sc SetConfigResponse) String() string {
 func (sc *SetConfigResponse) UnmarshalJSON(data []byte) error {
 	var err error
 	test := struct {
-		Code           *string  `json:"code,omitempty"`
-		Description    *string  `json:"description,omitempty"`
-		ExpectedValues []string `json:"expectedValues,omitempty"`
+		Success     bool    `json:"success"`
+		Description *string `json:"description,omitempty"`
 	}{}
 
 	//Verify that incoming string will unmarshal successfully
@@ -59,17 +52,10 @@ func (sc *SetConfigResponse) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
-	//If verified, copy the fields
-	if test.Code != nil {
-		sc.Code = *test.Code
-	}
+	sc.Success = test.Success
 
 	if test.Description != nil {
 		sc.Description = *test.Description
-	}
-
-	if test.ExpectedValues != nil {
-		sc.ExpectedValues = test.ExpectedValues
 	}
 
 	sc.isValidated, err = sc.Validate()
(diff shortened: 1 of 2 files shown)


responses/configuration/set.go
```diff
@@ -60,33 +54,26 @@ func (sc *SetConfigResponse) UnmarshalJSON(data []byte) error {
 	}
 
 	//If verified, copy the fields
-	if test.Code != nil {
-		sc.Code = *test.Code
+	if test.Success != false {
+		sc.Success = test.Success
```
michaelestrin: Is this assignment correct?

responses/configuration/set.go
```diff
@@ -60,33 +54,26 @@ func (sc *SetConfigResponse) UnmarshalJSON(data []byte) error {
...
 	sc.isValidated, err = sc.Validate()
 
 	return err
 }
 
 // Validate satisfies the Validator interface
 func (sc SetConfigResponse) Validate() (bool, error) {
-	if sc.Code == "" {
-		return false, models.NewErrContractInvalid(fmt.Sprintf("invalid Code %s", sc.Code))
+	if sc.Success != true {
```
michaelestrin: Is this correct?  Isn't a valid Success state false?

akramtexas: That is exactly what I had in mind when I had encoded the logical test initially, as follows...
`if sc.Success != true || sc.Success != false {`
...and the feedback I got from the IDE was that `expression can be simplified` (This appears to be a case where IDE feedback should be ignored).

michaelestrin: Validate()'s purpose, I think, is to determine whether or not the structure is in a valid state. 

Your code is looking at the SetConfigResponse.Success value -- which could either be true or false -- and only considering the structure to be valid when the Success is true.  I think the structure is valid when Success is false as well.

responses/configuration/set.go
```diff
@@ -8,26 +8,21 @@ import (
 
 //SetConfigResponse is for SMA to use when responding to an incoming request to PUT (i.e. UPDATE) a resource.
 type SetConfigResponse struct {
-	Code           string   `json:"code" yaml:"code,omitempty"`                     //HTTP code to return (e.g. 200, 500, etc.)
-	Description    string   `json:"description" yaml:"description,omitempty"`       //info allied with preceding HTTP code
-	ExpectedValues []string `json:"expectedValues" yaml:"expectedValues,omitempty"` //expected values to return
-	isValidated    bool     //internal member used for validation check
+	Success     bool   `json:"success"`               // HTTP success or failure
```
michaelestrin: Nitpick - not really HTTP success or failure; it's the success or failure of the set configuration request.

responses/configuration/set.go
```diff
@@ -8,26 +8,21 @@ import (
...
+	Description string `json:"description,omitempty"` // info allied with preceding HTTP code
```
michaelestrin: Nitpick - preceding field is not an HTTP code.

responses/configuration/set.go
```diff
@@ -8,26 +8,21 @@ import (
...
-	}
+		Success     bool    `json:"success"`
+		Description *string `json:"description,omitempty"`
+		isValidated bool
+	}{}
 
-	// Empty strings are null
-	if sc.Code != "" {
-		test.Code = &sc.Code
+	if sc.Success != false {
```
michaelestrin: While this code technically works, it isn't correct.  You should always assign `test.Success = sc.Success` regardless of the sc.Success value.

responses/configuration/set.go
```diff
@@ -60,17 +54,11 @@ func (sc *SetConfigResponse) UnmarshalJSON(data []byte) error {
...
-
 	if test.Description != nil {
 		sc.Description = *test.Description
 	}
-
-	if test.ExpectedValues != nil {
-		sc.ExpectedValues = test.ExpectedValues
-	}
+	// The field "Success" is equally valid when it holds a "true" or "false" value, so verification is irrelevant.
+	sc.Success = test.Success
```
michaelestrin: Nitpick - for consistency with the struct layout, I'd move this assignment before the `if test.Description...` line above.

responses/configuration/set.go
```diff
@@ -79,14 +67,11 @@ func (sc *SetConfigResponse) UnmarshalJSON(data []byte) error {
 
 // Validate satisfies the Validator interface
 func (sc SetConfigResponse) Validate() (bool, error) {
-	if sc.Code == "" {
-		return false, models.NewErrContractInvalid(fmt.Sprintf("invalid Code %s", sc.Code))
-	}
+	// The field "Success" is equally valid when it holds a "true" or "false" value, and a validation with a
+	// logical OR (i.e. if sc.Success != true || sc.Success != false) will be flagged in error by the strongly-typed
+	// Go language, so we don't validate that field. Validate only the field "Description"
```
michaelestrin: These three comment lines are unnecessary IMHO.

responses/configuration/set.go
```diff
@@ -60,17 +54,11 @@ func (sc *SetConfigResponse) UnmarshalJSON(data []byte) error {
...
-	}
-
 	if test.Description != nil {
 		sc.Description = *test.Description
 	}
-
-	if test.ExpectedValues != nil {
-		sc.ExpectedValues = test.ExpectedValues
-	}
+	// The field "Success" is equally valid when it holds a "true" or "false" value, so verification is irrelevant.
```
michaelestrin: This comment line is unnecessary IMHO.

responses/configuration/set_test.go
```diff
@@ -11,11 +11,9 @@ func TestGetValidation(t *testing.T) {
...
-		{"invalid - no parameter (Code, Description,ExpectedValues) proper", SetConfigResponse{Code: "", Description: "", ExpectedValues: nil}, true},
-		{"invalid - some parameters (Code, Description) proper, others not proper", SetConfigResponse{Code: "Logging.EnableRemote", Description: "false", ExpectedValues: nil}, true},
-		{"invalid - some parameters (Code, ExpectedValues) proper, others not proper", SetConfigResponse{Code: "Logging.EnableRemote", Description: "", ExpectedValues: []string{"abc", "def"}}, true},
-		{"invalid - some parameters (Description,ExpectedValues) proper, others not proper", SetConfigResponse{Code: "", Description: "true", ExpectedValues: []string{"abc", "def"}}, true},
+		{"valid   - all parameters (Success, Description) proper", SetConfigResponse{Success: true, Description: "Success"}, false},
+		{"valid   - all parameters (Success, Description) proper", SetConfigResponse{Success: false, Description: "Success"}, false},
+		{"invalid - one parameter (Success) proper, other not proper", SetConfigResponse{Success: true, Description: ""}, true},
```
michaelestrin: For completeness' sake, you might add an additional case where description is "" and success == false.

PR #157, opened 2019-09-25
Marshal empty objects as {}, simplify Marshal functions

### LINKED ISSUE #143: Several types String methods always include fields, even when empty
Follow up from #139 

The following types always include fields in their `String` method, even when empty:

- Action
- CallbackAlert
- CommandResponse
- DescribedObject
- Device
- Get
- IntervalAction
- Interval
- ProfileProperty
- ProvisionWatcher
- Put
- ResourceOperation
- Units
- ValueDescriptor

However, for some of these, an empty object is an invalid state. Should we implement expected String behavior for all of these, only the ones where an empty object is a valid state, or none of them?

brandonforster:
Fixes #143.

This PR will remove several superfluous unit test cases. I did not replace them because the functionality they sought to test is already covered by the `string()` tests for each of them. Every object includes a test asserting that an empty version of itself will be marshalled as `{}`.

I removed a lot of logic from the custom marshalling functions. I would have removed the marshal functions themselves for many of these objects, but I wanted to preserve backward compatibility. That is something that could be addressed in a follow on PR.


models/transmission.go
```diff
@@ -35,28 +36,40 @@ type Transmission struct {
...
+		Channel:      &t.Channel,
 		Status:       t.Status,
 		ResendCount:  t.ResendCount,
 		Records:      t.Records,
 	}
-	// Empty strings are null
-	if t.ID != "" {
-		test.ID = &t.ID
+
+	// this is a horrible, no good hack but is necessary because a ResendCount of 0 trips up the omitempty tag
```
brandonforster: I want to call out this and ask if anyone has a better idea for how to accomplish this.

models/transmission.go
```diff
@@ -35,28 +36,40 @@ type Transmission struct {
...
+	// if we don't use omitempty, then an empty object always has a ResendCount included
+	// if we do use omitempty, then a ResendCount of 0 is not included in the object when it should be
+	if reflect.DeepEqual(t, Transmission{}) {
+		return []byte("{}"), nil
```
brandonforster: Calling this out after rebase.

models/deviceprofile.go
```diff
@@ -22,62 +22,17 @@ import (
...
 	DescribedObject `yaml:",inline"`
-	Id              string            `json:"id" yaml:"id,omitempty"`
-	Name            string            `json:"name" yaml:"name,omitempty"`                 // Non-database identifier (must be unique)
-	Manufacturer    string            `json:"manufacturer" yaml:"manufacturer,omitempty"` // Manufacturer of the device
-	Model           string            `json:"model" yaml:"model,omitempty"`               // Model of the device
-	Labels          []string          `json:"labels" yaml:"labels,flow,omitempty"`        // Labels used to search for groups of profiles
-	DeviceResources []DeviceResource  `json:"deviceResources" yaml:"deviceResources,omitempty"`
-	DeviceCommands  []ProfileResource `json:"deviceCommands" yaml:"deviceCommands,omitempty"`
-	CoreCommands    []Command         `json:"coreCommands" yaml:"coreCommands,omitempty"` // List of commands to Get/Put information for devices associated with this profile
+	Id              string            `json:"id,omitempty" yaml:"id,omitempty"`
```
michaelestrin: Should you be adding omitempty to the yaml tag?

Calling out once -- occurs in multiple places.

models/deviceresource.go
```diff
@@ -30,29 +31,24 @@ type DeviceResource struct {
...
-		Description *string           `json:"description,omitempty"`
-		Name        *string           `json:"name,omitempty"`
-		Tag         *string           `json:"tag,omitempty"`
-		Properties  ProfileProperty   `json:"properties"`
-		Attributes  map[string]string `json:"attributes,omitempty"`
+		Description string             `json:"description,omitempty"`
+		Name        string             `json:"name,omitempty"`
+		Tag         string             `json:"tag,omitempty"`
+		Properties  *ProfileProperty   `json:"properties,omitempty"`
+		Attributes  *map[string]string `json:"attributes,omitempty"`
```
michaelestrin: Why was this converted to a pointer?  omitempty tag will not include a map if it's empty.

models/event.go
```diff
@@ -21,13 +21,13 @@ import (
...
-	Pushed      int64     `json:"pushed" codec:"pushed,omitempty"`     // Pushed is a timestamp indicating when the event was exported. If unexported, the value is zero.
-	Device      string    `json:"device" codec:"device,omitempty"`     // Device identifies the source of the event, can be a device name or id. Usually the device name.
-	Created     int64     `json:"created" codec:"created,omitempty"`   // Created is a timestamp indicating when the event was created.
-	Modified    int64     `json:"modified" codec:"modified,omitempty"` // Modified is a timestamp indicating when the event was last modified.
-	Origin      int64     `json:"origin" codec:"origin,omitempty"`     // Origin is a timestamp that can communicate the time of the original reading, prior to event creation
-	Readings    []Reading `json:"readings" codec:"readings,omitempty"` // Readings will contain zero to many entries for the associated readings of a given event.
+	ID          string    `json:"id,omitempty" codec:"id,omitempty"`             // ID uniquely identifies an event, for example a UUID
```
michaelestrin: Should you be adding omitempty to codec tag?

Calling out once -- occurs in multiple places.

brandonforster:
Closing to split into multiple.

models/event.go
brandonforster: No. Copy paste error. Good catch, thanks.

models/callbackalert_test.go
```diff
@@ -15,34 +15,39 @@
...
 	"reflect"
 	"testing"
 )
 
 var TestCallbackAlert = CallbackAlert{"DEVICE", "1234"}
+var TestEmptyCallbackAlert = CallbackAlert{}
 
 func TestCallbackAlert_MarshalJSON(t *testing.T) {
 	var testNoIDCallbackAlert = CallbackAlert{"DEVICE", ""}
 	var testCallbackAlertBytes = []byte(TestCallbackAlert.String())
```
brandonforster: @jdharms

PR #163, opened 2019-10-08
Updated Addressable model to correctly marshal an empty object.

brandonforster:
This is PR 2 of approximately 35 addressing issue #143.

### DIFF
--- a/models/addressable_test.go
+++ b/models/addressable_test.go
@@ -22,6 +22,18 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 )
 
+const (
+	testAddrName  = "TEST_ADDR.NAME"
+	testProtocol  = "http"
+	testMethod    = "Get"
+	testAddress   = "localhost"
+	testPort      = 48089
+	testPublisher = "TEST_PUB"
+	testUser      = "edgexer"
+	testPassword  = "password"
+	testTopic     = "device_topic"
+)
+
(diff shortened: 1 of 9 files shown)


models/addressable_test.go
```diff
@@ -36,16 +36,13 @@ var TestAddressable = Addressable{Timestamps: testTimestamps, Name: testAddrName
...
-	var resultEmptyBytes = []byte(EmptyAddressable.String())
 	tests := []struct {
 		name    string
 		a       Addressable
 		want    []byte
 		wantErr bool
 	}{
-		{"successful marshal", TestAddressable, resultTestBytes, false},
-		{"successful empty marshal", EmptyAddressable, resultEmptyBytes, false},
+		{"successful empty marshal", EmptyAddressable, []byte(TestEmptyJSON), false},
```
michaelestrin: Missed this on first review.  Pull TestEmptyJSON (which is now being used in multiple model tests ) into own file.

brandonforster: Do you want all files to have only local test variables, or do you want to limit this scope to this constant?

michaelestrin: Preference is to not have one test take a dependency on constants or variables defined in a separate test (but within the same package).  Extract the commonly used dependencies defined in a single test file into their own file/module.

michaelestrin: > Preference is to not have one test take a dependency on constants or variables defined in a separate test (but within the same package). Extract the commonly used dependencies defined in a single test file into their own file/module.

You moved constants from test file that are only currently being used by the test.  The ask was to move those that were being used across tests, not to move all constants into a common file/module.

I think I'm missing something...

PR #167, opened 2019-10-15
Update models to serialize zero values as empty JSON objects, part 2

jdharms:
Part of the work to address #143

Addresses the following models:

CallbackAlert
Channel
Command
CommandResponse
Device
DeviceProfile
DeviceService

### DIFF
--- a/models/callbackalert_test.go
+++ b/models/callbackalert_test.go
@@ -15,34 +15,33 @@
 package models
 
 import (
+	"encoding/json"
 	"reflect"
 	"testing"
 )
 
 var TestCallbackAlert = CallbackAlert{"DEVICE", "1234"}
+var TestEmptyCallbackAlert = CallbackAlert{}
 
 func TestCallbackAlert_MarshalJSON(t *testing.T) {
-	var testNoIDCallbackAlert = CallbackAlert{"DEVICE", ""}
-	var testCallbackAlertBytes = []byte(TestCallbackAlert.String())
-	var testNoIDCallbackAlertBytes = []byte(testNoIDCallbackAlert.String())
 	tests := []struct {
 		name    string
-		ca      CallbackAlert
+		ca      interface{}
 		want    []byte
 		wantErr bool
 	}{
-		{"successful marshal of callback", TestCallbackAlert, testCallbackAlertBytes, false},
-		{"successful marshal of no id callback", testNoIDCallbackAlert, testNoIDCallbackAlertBytes, false},
+		{"successful marshal of empty callback", TestEmptyCallbackAlert, []byte(testEmptyJSON), false},
+		{"unsuccessful marshal", make(chan int), nil, true},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := tt.ca.MarshalJSON()
+			got, err := json.Marshal(tt.ca)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("CallbackAlert.MarshalJSON() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
 			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("CallbackAlert.MarshalJSON() = %v, want %v", got, tt.want)
+				t.Errorf("CallbackAlert.MarshalJSON() = %v, want %v", string(got), string(tt.want))
 			}
 		})
 	}
@@ -54,7 +53,8 @@ func TestCallbackAlert_String(t *testing.T) {
 		ca   CallbackAlert
 		want string
 	}{
-		{"callback alert to string", TestCallbackAlert, "{\"type\":\"DEVICE\",\"id\":\"1234\"}"},
+		{"successful callback alert to string", TestCallbackAlert, "{\"type\":\"DEVICE\",\"id\":\"1234\"}"},
+		{"successful empty callback alert to string", TestEmptyCallbackAlert, testEmptyJSON},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/models/command.go
+++ b/models/command.go
@@ -33,23 +33,16 @@ type Command struct {
 func (c Command) MarshalJSON() ([]byte, error) {
 	test := struct {
 		Timestamps
-		Id   *string `json:"id,omitempty"`
-		Name *string `json:"name,omitempty"` // Command name (unique on the profile)
-		Get  *Get    `json:"get,omitempty"`  // Get Command
-		Put  *Put    `json:"put,omitempty"`  // Put Command
+		Id   string `json:"id,omitempty"`
+		Name string `json:"name,omitempty"` // Command name (unique on the profile)
+		Get  *Get   `json:"get,omitempty"`  // Get Command
+		Put  *Put   `json:"put,omitempty"`  // Put Command
 	}{
 		Timestamps: c.Timestamps,
+		Id:         c.Id,
+		Name:       c.Name,
 		Get:        &c.Get,
 		Put:        &c.Put,
-	}
-
-	if c.Id != "" {
-		test.Id = &c.Id
-	}
-
-	// Make empty strings null
-	if c.Name != "" {
-		test.Name = &c.Name
 	}
 
 	// Make empty structs nil pointers so they aren't marshaled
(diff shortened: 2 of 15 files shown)


models/callbackalert_test.go
```diff
@@ -15,34 +15,33 @@
...
 		name    string
-		ca      CallbackAlert
+		ca      interface{}
 		want    []byte
 		wantErr bool
 	}{
-		{"successful marshal of callback", TestCallbackAlert, testCallbackAlertBytes, false},
-		{"successful marshal of no id callback", testNoIDCallbackAlert, testNoIDCallbackAlertBytes, false},
+		{"successful marshal of empty callback", TestEmptyCallbackAlert, []byte(testEmptyJSON), false},
+		{"unsuccessful marshal", make(chan int), nil, true},
```
brandonforster: Fantastic addition.

models/deviceprofile.go
```diff
@@ -37,34 +37,26 @@ type DeviceProfile struct {
...
-	}
-	if dp.Manufacturer != "" {
-		test.Manufacturer = &dp.Manufacturer
-	}
-	if dp.Model != "" {
-		test.Model = &dp.Model
-	}
-
 	// Empty arrays are null
+	if len(dp.Labels) > 0 {
```
brandonforster: Arrays don't need custom initialization.

jdharms: Sure enough, thanks.

models/command.go
```diff
@@ -33,25 +33,18 @@ type Command struct {
 func (c Command) MarshalJSON() ([]byte, error) {
 	test := struct {
 		Timestamps
-		Id   *string `json:"id,omitempty"`
-		Name *string `json:"name,omitempty"` // Command name (unique on the profile)
-		Get  *Get    `json:"get,omitempty"`  // Get Command
-		Put  *Put    `json:"put,omitempty"`  // Put Command
+		Id   string `json:"id,omitempty"`
```
michaelestrin: Can this be simplified in a similar manner to https://github.com/edgexfoundry/go-mod-core-contracts/pull/163#pullrequestreview-299392862?

i.e.

```
type commandCommon struct {
	Timestamps `yaml:",inline"`
	Id         string `json:"id,omitempty" yaml:"id,omitempty"`     // Id is a unique identifier, such as a UUID
	Name       string `json:"name,omitempty" yaml:"name,omitempty"` // Command name (unique on the profile)
}

type commonAlias commandCommon

// Command defines a specific read/write operation targeting a device
type Command struct {
	commandCommon
	Get         Get  `json:"get" yaml:"get,omitempty"` // Get Command
	Put         Put  `json:"put" yaml:"put,omitempty"` // Put Command
	isValidated bool // internal member used for validation check
}

type commandAlias Command

// MarshalJSON implements the Marshaler interface. Empty strings will be null.
func (c Command) MarshalJSON() ([]byte, error) {
	alias := struct {
		commonAlias
		Get *Get `json:"get,omitempty"` // Get Command
		Put *Put `json:"put,omitempty"` // Put Command
	}{
		commonAlias: commonAlias(c.commandCommon),
	}
	// Make empty structs nil pointers so they aren't marshaled
	if !reflect.DeepEqual(c.Get, Get{}) {
		alias.Get = &c.Get
	}
	if !reflect.DeepEqual(c.Put, Put{}) {
		alias.Put = &c.Put
	}
	return json.Marshal(alias)
}

// UnmarshalJSON implements the Unmarshaler interface for the Command type
func (c *Command) UnmarshalJSON(data []byte) error {
	var err error
	var alias commandAlias
	if err = json.Unmarshal(data, alias); err != nil {
		return err
	}
	*c = Command(alias)
	c.isValidated, err = c.Validate()
	return err
}
```

Note impact on corresponding tests...

```
var TestCommand = Command{commandCommon: commandCommon{Timestamps: testTimestamps, Id: TestId, Name: testCommandName,}, Get: TestGet, Put: TestPut}
var TestCommandGetOnly = Command{commandCommon: commandCommon{Timestamps: testTimestamps, Id: TestId, Name: testCommandName,}, Get: TestGet}
var TestCommandPutOnly = Command{commandCommon: commandCommon{Timestamps: testTimestamps, Id: TestId, Name: testCommandName,}, Put: TestPut}
```

This approach eliminates as much struct field duplication as possible given we can't change the signature of Get and Put in Command from values to pointers.

Only calling out this instance; there may be others.

jdharms:
Updated to address feedback

models/command.go
michaelestrin: Which is why it's package private.

I'm not clear on what you think is idomatic (or not) about the suggested approach.

It's interesting that you chose not to simplify UnmarshalJSON as suggested.  The code could be simplified to not have yet another anonymous struct (that needs to be kept in sync with the others) and has nothing to do with the statements you've made above.

My suggested approach is consistent with feedback on other PRs related to resolving the common parent issue.  If we choose not to go this route, we'll need to be consistent across all PRs for the issue.

Again, I'll ask @tsconn23 to weigh in to see how he wants to proceed.

brandonforster: `UnmarshalJSON` is out of scope of these issues and I made the changes there because I didn't have an objection to it, but I don't think it should be used to buttress your argument.

michaelestrin: Appreciate your input @brandonforster.  Again, I'll ask @tsconn23 to weigh in to see how he wants to proceed.

jdharms: > Which is why it's package private.

>I'm not clear on what you think is idomatic (or not) about the suggested approach.

Sure, I am saying it is not idiomatic to have fields embedded instead of as public members of the struct to clean up some serialization code elsewhere.  Clients being unable to instantiate this struct doesn't change that, but I can what was confusing there.

michaelestrin: @jdharms -- I can call anything idiomatic (or not).  Can you supply a reference?

I'm disengaging from this conversation.  @tsconn23 will need to review and provide direction.  Related conversation here -- https://github.com/edgexfoundry/go-mod-core-contracts/pull/168#discussion_r335604334

tsconn23: Here are my thoughts 

I don't think the introduction of the `commandCommon` is necessarily un-idiomatic, but I personally don't think it adds much value. From what I can see, it's basically to encapsulate properties contained in the alias type and the exported type. It avoids duplication of a handful of lines and won't be used anywhere else in the package. Therefore I personally don't see it as a necessary change to approve the PR.

Let's not lose focus here. We have a handful of days to get these changes completed for inclusion in the Fuji release. The optimal goal is a reduction in the amount of JSON being passed over the wire, via improved handling of "empty" types. I am much less concerned with wholesale refactoring to eliminate a minor percentage of duplicated lines.

Having said that, there's now the follow up issue of this pattern being in some models but not in others. I personally am not going to lose a whole lot of sleep over this given the amount of change this repo is going to see in the Geneva work cycle (per Core WG call this morning). However in the future, my suggestion would be that if there are design choices which have been made and agreed to (either explicitly or implicitly) by resources assigned to a task, and then additional resources come online to complete the effort, those additional resources are beholden to adhere to the agreed upon patterns. 

My recommendation on this PR is that if it accomplishes the desired goal and has been successfully tested via an integrated flow between an example device service and edgex-go, as well as any relevant blackbox tests, approve it.

tsconn23:
@brandonforster @jdharms I just discussed the following with Brandon on a call. Here's what I'd like to be confident of w/r/t the level of testing each of these PRs receives. The last thing I want is for us to merge this whole stream, then next Tuesday on the day before code freeze we consume the updated module in edgex-go and find ourselves faced with 100 blackbox regressions.

```
(local) 2          (local) 1          (pr_176)         1 (local)  
device-random --> device-sdk-go --> core-contracts <-- edgex-go
```

To elaborate on this, 
1.) Get latest on the feature branch of pr 176
2.) Point the relevant `go-mod-core-contracts` reference in the go.mod of both the device-sdk-go and edgex-go to the pr 176 feature branch.
3.) Build both of those
4.) Now point device-random's reference to `device-sdk-go` to your local copy via the go.mod
5.) Build that
6.) At this point you should be able to run an end to end flow whereby events from device-random are ingested into edgex-go. You should also be able to actuate commands from core-command to the device service and see those work.
7.) Once that flow has been established, run all the blackbox tests, ensuring you have the latest test data and test collections.

This has to be done for each PR. Once the above testing plan has been completed, the PR can be merged. Please let me know if you have any questions.

brandonforster:
Closing to merge into a feature branch.

PR #168, opened 2019-10-15
Update models to serialize zero values as empty JSON objects, part 3

brandonforster:
Part 3 of 6 for #143

This work includes: sma_operation, subscription, timestamps, transmission, transmission_record, units, value-descriptor

### DIFF
--- a/models/transmission.go
+++ b/models/transmission.go
@@ -17,6 +17,7 @@ package models
 
 import (
 	"encoding/json"
+	"reflect"
 )
 
 type Transmission struct {
@@ -31,33 +32,42 @@ type Transmission struct {
 	isValidated  bool
 }
 
-// Custom marshaling to make empty strings null
+// Marshal returns a JSON encoded byte array representation of the model
 func (t Transmission) MarshalJSON() ([]byte, error) {
-	test := struct {
+	alias := struct {
 		Timestamps
-		ID           *string              `json:"id"`
-		Notification Notification         `json:"notification,omitempty"`
-		Receiver     *string              `json:"receiver,omitempty"`
-		Channel      Channel              `json:"channel,omitempty"`
+		ID           string               `json:"id,omitempty"`
+		Notification *Notification        `json:"notification,omitempty"`
+		Receiver     string               `json:"receiver,omitempty"`
+		Channel      *Channel             `json:"channel,omitempty"`
 		Status       TransmissionStatus   `json:"status,omitempty"`
-		ResendCount  int                  `json:"resendcount"`
+		ResendCount  *int                 `json:"resendcount,omitempty"`
 		Records      []TransmissionRecord `json:"records,omitempty"`
 	}{
 		Timestamps:   t.Timestamps,
-		Notification: t.Notification,
-		Channel:      t.Channel,
+		ID:           t.ID,
+		Notification: &t.Notification,
+		Receiver:     t.Receiver,
+		Channel:      &t.Channel,
 		Status:       t.Status,
-		ResendCount:  t.ResendCount,
+		ResendCount:  &t.ResendCount,
 		Records:      t.Records,
 	}
-	// Empty strings are null
-	if t.ID != "" {
-		test.ID = &t.ID
+
+	// if we don't use omitempty, then an empty object always has a ResendCount included
+	// if we do use omitempty, then a ResendCount of 0 is not included in the object when it should be
+	if reflect.DeepEqual(t, Transmission{}) {
+		alias.ResendCount = nil
 	}
-	if t.Receiver != "" {
-		test.Receiver = &t.Receiver
+	// do not marshal empty member objects
+	if reflect.DeepEqual(t.Notification, Notification{}) {
+		alias.Notification = nil
 	}
-	return json.Marshal(test)
+	if reflect.DeepEqual(t.Channel, Channel{}) {
+		alias.Channel = nil
+	}
+
+	return json.Marshal(alias)
 }
 
 // UnmarshalJSON implements the Unmarshaler interface for the Transmission type
@@ -78,7 +88,6 @@ func (t *Transmission) UnmarshalJSON(data []byte) error {
 	if err = json.Unmarshal(data, &a); err != nil {
 		return err
 	}
-
 	// Nillable fields
 	if a.ID != nil {
 		t.ID = *a.ID
@@ -128,9 +137,7 @@ func (t Transmission) Validate() (bool, error) {
 	return t.isValidated, nil
 }
 
-/*
- * To String function for Transmission Struct
- */
+// String returns a JSON encoded string representation of the model
 func (t Transmission) String() string {
 	out, err := json.Marshal(t)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


models/transmission.go
```diff
@@ -31,32 +32,44 @@ type Transmission struct {
 	isValidated  bool
 }
 
-// Custom marshaling to make empty strings null
+// Marshal returns a JSON encoded byte array representation of the model
```
michaelestrin: Can this be simplified in a similar manner to https://github.com/edgexfoundry/go-mod-core-contracts/pull/163#pullrequestreview-299392862?

i.e.

```
type transmissionCommon struct {
	Timestamps
	ID           string               `json:"id,omitempty"`
	Receiver     string               `json:"receiver,omitempty"`
	Status       TransmissionStatus   `json:"status,omitempty"`
	Records      []TransmissionRecord `json:"records,omitempty"`
}

type transmissionCommonAlias transmissionCommon

type Transmission struct {
	transmissionCommon
	Notification Notification         `json:"notification"`
	Channel      Channel              `json:"channel,omitempty"`
	ResendCount  int                  `json:"resendcount"`
	isValidated  bool
}

type transmissionAlias Transmission

// Marshal returns a JSON encoded byte array representation of the model
func (t Transmission) MarshalJSON() ([]byte, error) {
	alias := struct {
		transmissionCommonAlias
		Notification *Notification         `json:"notification,omitempty"`
		Channel      *Channel              `json:"channel,omitempty"`
		ResendCount  *int                  `json:"resendcount,omitempty"`
	}{
		transmissionCommonAlias: transmissionCommonAlias(t.transmissionCommon),
	}
	if !reflect.DeepEqual(t, Transmission{}) {
		alias.ResendCount = &t.ResendCount
	}
	// do not marshal empty member objects
	if !reflect.DeepEqual(t.Notification, Notification{}) {
		alias.Notification = &t.Notification
	}
	if !reflect.DeepEqual(t.Channel, Channel{}) {
		alias.Channel = &t.Channel
	}
	return json.Marshal(alias)
}

// UnmarshalJSON implements the Unmarshaler interface for the Transmission type
func (t *Transmission) UnmarshalJSON(data []byte) error {
	var err error
	var alias transmissionAlias
	if err = json.Unmarshal(data, &alias); err != nil {
		return err
	}
	*t = Transmission(alias)
	t.isValidated, err = t.Validate()
	return err
}
```

Note impact on corresponding tests...

```
var TestTransmission = Transmission{
	transmissionCommon: transmissionCommon{
		Timestamps: testTimestamps,
		Receiver: "test receiver",
		Status: TransmissionStatus(Sent),
		Records: []TransmissionRecord{TestSentTransRecord},
	},
	Notification: TestNotification,
	Channel: Channel{
		Type: ChannelType(Email),
		MailAddresses: []string{"jpwhite_mn@yahoo.com", "james_white2@dell.com"},
	},
	ResendCount: 0,
}
...
		{"test string of transmission", &TestTransmission, "{\"created\":123,\"modified\":123,\"origin\":123,\"receiver\":\"test receiver\",\"status\":\"SENT\",\"records\":[{\"status\":\"SENT\",\"response\":\"ok\",\"sent\":123}],\"notification\":{\"created\":123,\"modified\":123,\"id\":\"" + TestNotificationID + "\",\"slug\":\"test slug\",\"sender\":\"test sender\",\"category\":\"SECURITY\",\"severity\":\"CRITICAL\",\"content\":\"test content\",\"description\":\"test description\",\"status\":\"NEW\",\"labels\":[\"label1\",\"labe2\"],\"contenttype\":\"text/plain\"},\"channel\":{\"type\":\"EMAIL\",\"mailAddresses\":[\"jpwhite_mn@yahoo.com\",\"james_white2@dell.com\"]},\"resendcount\":0}"},

```

This approach eliminates as much struct field duplication as possible and gets rid of the hack returning "{}" string literal.

Only calling out this instance; there may be others.

models/transmission_record_test.go
```diff
@@ -25,37 +24,6 @@ var TestSentTransRecord = TransmissionRecord{Response: "ok", Sent: 123, Status:
 var TestEmptyRespTransRecord = TransmissionRecord{Sent: 123, Status: TransmissionStatus(Sent)}
 var TestEmptyTransRecord = TransmissionRecord{}
 
-func TestTransmissionRecord_MarshalJSON(t *testing.T) {
```
michaelestrin: Why remove the tests?  Are they not still valid?

Only calling out this instance; there may be others.

brandonforster: The method no longer exists and is functionally duplicated by `_String()` anyway.

It does not need a test if it does not exist. We do not need to test the core library.

michaelestrin: Yes, and...  ...it begets the question of why test the MarshalJSON methods directly at all if we've got a separate method that marshals the structure and returns JSON that has its own tests.

MarshalJSON -- while part of the receiver's public interface -- is magical and isn't intended to be called directly.

michaelestrin:
@brandonforster please re-request a review if/when you've finished making changes and are ready for me to do another review.

models/transmission_record_test.go
brandonforster: > the question of why test the MarshalJSON methods directly at all

I absolutely agree. This issue's scope is constantly creeping so I don't want to delete all the MarshalJSON tests, but they are pretty useless on the whole.

> MarshalJSON -- while part of the receiver's public interface -- is magical and isn't intended to be called directly.

Again, `transmission` doesn't have a MarshalJSON method. I absolutely refuse to have a test for a method we don't have. Should we also have tests for length methods? Typeofs?

brandonforster:
@michaelestrin You have not requested any changes.

models/transmission_record_test.go
brandonforster: Sorry, I was confused.

`transmission_record` no longer has a custom marshaller. Since it does not have any custom logic, it does not need any tests. We cannot test a method if it doesn't exist.

The MarshalJSON methods on the whole are pretty useless, but I think deleting all of them is a bad idea when they aren't hurting anything.

michaelestrin: I think I'm not communicating well.

I meant to say that if we've got tests around MarshalJSON methods and tests around String() methods (whose implementation calls json.Marshal which subsequently magically calls the aforementioned MarshalJSON method), it seems like we've got duplication in the tests and the MarshalJSON tests should be removed (regardless of whether or not a MarshalJSON method exists for the receiver being tested).

As for deleting MarshalJSON methods, we agreed based on Trevor's input to remove those that aren't doing custom marshalling (i.e. have no reason to exist).  I'm not sure if you're questioning the wisdom of that or ...?  

I know there's a sense of scope creep here.  I'd prefer to frame it as following [the Boy Scout rule](https://medium.com/@biratkirat/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385).  The existing model code reeks in places and it makes sense -- given that we're actively modifying it to address this issue -- to spend a little extra energy to clean it up a bit.

brandonforster:
UnmarshalJSON is outside the scope of this work.

brandonforster:
Closing to merge into a feature branch.

PR #177, opened 2019-10-18
Feature merge: Update models to serialize zero values as empty JSON objects

### LINKED ISSUE #143: Several types String methods always include fields, even when empty

brandonforster:
Fix #143


jdharms:
Superseded by #180

PR #182, opened 2019-10-24
Create new LoggingClient factory that does not check logTarget.

### LINKED ISSUE #181: "logTarget cannot be blank" error for all services consuming logging
This issue tracks the bug described by /edgexfoundry/edgex-go/issues/1989

brandonforster:
Fixes #181 

### DIFF
--- a/clients/logger/logger.go
+++ b/clients/logger/logger.go
@@ -67,10 +67,6 @@ type fileWriter struct {
 
 // NewClient creates an instance of LoggingClient
 func NewClient(owningServiceName string, isRemote bool, logTarget string, logLevel string) LoggingClient {
-	if !IsValidLogLevel(logLevel) {
-		logLevel = models.InfoLog
-	}
-
 	lc := newClient(owningServiceName, isRemote, logTarget, logLevel)
 
 	if logTarget == "" {
@@ -82,17 +78,15 @@ func NewClient(owningServiceName string, isRemote bool, logTarget string, logLev
 
 // NewClientStdOut creates an instance of LoggingClient that expects to log to stdout and does not check logTarget
 func NewClientStdOut(owningServiceName string, isRemote bool, logLevel string) LoggingClient {
-	if !IsValidLogLevel(logLevel) {
-		logLevel = models.InfoLog
-	}
-
-	lc := newClient(owningServiceName, isRemote, "", logLevel)
-
-	return lc
+	return newClient(owningServiceName, isRemote, "", logLevel)
 }
 
 // newClient is the implementation of the logic required for the factory functions
 func newClient(owningServiceName string, isRemote bool, logTarget string, logLevel string) edgeXLogger {
+	if !IsValidLogLevel(logLevel) {
+		logLevel = models.InfoLog
+	}
+
 	// Set up logging client
 	lc := edgeXLogger{
 		owningServiceName: owningServiceName,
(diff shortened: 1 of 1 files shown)


clients/logger/logger.go
```diff
@@ -67,6 +67,17 @@ type fileWriter struct {
...
+
+	if logTarget == "" {
+		lc.Error("logTarget cannot be blank, using stdout only")
+	}
+
+	return lc
+}
+
+// NewClientStdOut creates an instance of LoggingClient that expects to log to stdout and does not check logTarget
+func NewClientStdOut(owningServiceName string, isRemote bool, logTarget string, logLevel string) LoggingClient {
```
michaelestrin: Refactor your implementation so NewClientStdOut doesn't require a logTarget parameter.

brandonforster: Disregard.

michaelestrin: It's not an either/or.

Your factory functions are implementation.  They can leverage a new public method (which is not part of the LoggingClient contract) on the receiver to return what you need.

Let me know if you want to pair code this.

brandonforster: Please see the latest commit.

clients/logger/logger.go
```diff
@@ -152,6 +130,35 @@ func newFileWriter(logTarget string) (io.Writer, error) {
 	return &fileWriter, nil
 }
 
+func (lc *edgeXLogger) initLogger(owningServiceName string, isRemote bool, logTarget, logLevel string) {
```
michaelestrin: Violates single responsibility principle.

Your factory functions are implementation. They can leverage a new public method (which is not part of the LoggingClient contract) on the receiver to return what you need.

Let me know if you want to pair code this.

clients/logger/logger.go
michaelestrin: I did.  Commented there.

clients/logger/logger.go
brandonforster: I disagree that it violates the SRP. This is a file that creates loggers. After my changes are merged, it will be a file that creates loggers. I'm interested to hear your take on why it isn't.

Further, I think your suggestion is adding a lot of complexity to a relatively straightforward solution to a small problem. What is the benefit of adding a new exported method? Who do we expect to consume that method? What is gained by doing this? With my proposed solution, we have one new exported function, one new unexported method, it is clear what the exported function does and its use case versus the existing exported function, and the existing exported function does not have any changed behavior. With the addition of the unexported method, there is no code duplication to worry about.

Unless I'm misunderstanding and you're suggesting we use a different name other than `NewClient` because that is part of the defined interface, then, sure. I don't have a problem with something like that.

brandonforster: Resolved on call.

clients/logger/logger.go
```diff
@@ -71,6 +71,28 @@ func NewClient(owningServiceName string, isRemote bool, logTarget string, logLev
...
+
+	return lc
+}
+
+// NewClientStdOut creates an instance of LoggingClient that expects to log to stdout and does not check logTarget
+func NewClientStdOut(owningServiceName string, isRemote bool, logLevel string) LoggingClient {
+	if !IsValidLogLevel(logLevel) {
+		logLevel = models.InfoLog
+	}
+
```
michaelestrin: Code is same in NewClient.  Both instances should be extracted into the common newClient().

PR #190, opened 2019-12-31
Add minimum supported Go version 1.12 to go.mod.

### LINKED ISSUE #161: Add minimum supported Go version to go.mod
Locally, I've upgraded to Go 1.13 to test it out. The project hasn't officially rev'ed to that version yet. One of the things I notice is that when I do a `make test` on a module, an entry indicating go 1.13 as the minimum supported version is put into the go.mod. According to the following article, we should have started seeing this behavior with Go 1.12, but I don't recall that being the case

https://golang.org/doc/go1.12#modules

I wouldn't want this to be checked in as go 1.13 until the project moves to that version. I've tested compilation of the edgex-go services locally with the version directive set to go 1.12 and everything works as far as I can tell. Again, this directive is only indicating a minimum supported version.

I think we should do this fairly soon as we could easily get contributions from people who've upgraded to Go 1.13 and who might miss this small edit.

akramtexas:
Fix #161 

Locally, I upgraded to Go `1.12`, and tested it out via `Makefile`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,3 +8,5 @@ require (
 	github.com/stretchr/testify v1.3.0
 	github.com/ugorji/go v1.1.4
 )
+
+go 1.12


akramtexas:
Recheck

akramtexas:
Recheck

PR #195, opened 2020-01-08
Refactor Client Code, part 1 of several

brandonforster:
This PR fixes https://github.com/edgexfoundry/edgex-go/issues/2237.

It is the first part of several working on #196.

While addressing the root cause of that issue, it was realized that the bug was not contained solely to the `device` endpoint, nor was it in command. The problem lies in how every endpoint that is exposed for cross service communication initializes its URL field.

This PR addresses the issue on metadata's `device` client, fixing 2237. It also lays out the strategy that will eventually be applied to every endpoint and the structures and objects that will be used to accomplish that. They were not added to this PR for the sake of review. Any feedback on this PR will be used to inform the future PRs.

To test, actuate the endpoints in the metadata device client. The blackbox tests for command should work for that.

### DIFF
--- a/clients/interfaces/builder.go
+++ b/clients/interfaces/builder.go
@@ -12,16 +12,15 @@
  * the License.
  *******************************************************************************/
 
-package client
+package interfaces
 
-import (
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
-)
-
-func Factory(params types.EndpointParams, m interfaces.Endpointer) interfaces.ClientBuilder {
-	if params.UseRegistry {
-		return newConsul(params, m, 10)
-	}
-	return newLocal(params)
+// RestClientBuilder is the interface for types that need to define some way to retrieve state information about services.
+// This information can be anything that must be determined at runtime, whether it is unknown or simply not yet known.
+type RestClientBuilder interface {
+	// URLPrefix returns the URL base path (or root) of a service.
+	// This is the common root of all REST calls to the service,
+	// and is defined on a per service (rather than per endpoint) basis.
+	// URLPrefix returns the root URL for REST calls to the service if it was able to retrieve that URL;
+	// it returns an error otherwise.
+	URLPrefix() (string, error)
 }
--- a/clients/metadata/device.go
+++ b/clients/metadata/device.go
@@ -21,6 +21,7 @@ import (
 	"strconv"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/common"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
@@ -75,12 +76,12 @@ type DeviceClient interface {
 }
 
 type deviceRestClient struct {
-	endpointer *clients.EndpointerClient
+	client *common.Client
 }
 
 // NewDeviceClient creates an instance of DeviceClient
 func NewDeviceClient(params types.EndpointParams, m interfaces.Endpointer) DeviceClient {
-	return &deviceRestClient{endpointer: clients.NewEndpointerClient(params, m)}
+	return &deviceRestClient{client: common.NewClient(params, m)}
 }
 
 // Helper method to request and decode a device
@@ -108,7 +109,7 @@ func (d *deviceRestClient) requestDeviceSlice(url string, ctx context.Context) (
 }
 
 func (d *deviceRestClient) CheckForDevice(token string, ctx context.Context) (models.Device, error) {
-	deviceURL, err := d.endpointer.URL(10)
+	deviceURL, err := d.client.URL(10)
 	if err != nil {
 		return models.Device{}, err
 	}
@@ -117,7 +118,7 @@ func (d *deviceRestClient) CheckForDevice(token string, ctx context.Context) (mo
 }
 
 func (d *deviceRestClient) Device(id string, ctx context.Context) (models.Device, error) {
-	deviceURL, err := d.endpointer.URL(10)
+	deviceURL, err := d.client.URL(10)
 	if err != nil {
 		return models.Device{}, err
 	}
@@ -126,7 +127,7 @@ func (d *deviceRestClient) Device(id string, ctx context.Context) (models.Device
 }
 
 func (d *deviceRestClient) Devices(ctx context.Context) ([]models.Device, error) {
-	deviceURL, err := d.endpointer.URL(10)
+	deviceURL, err := d.client.URL(10)
 	if err != nil {
 		return nil, err
 	}
@@ -135,7 +136,7 @@ func (d *deviceRestClient) Devices(ctx context.Context) ([]models.Device, error)
 }
 
 func (d *deviceRestClient) DeviceForName(name string, ctx context.Context) (models.Device, error) {
-	deviceURL, err := d.endpointer.URL(10)
+	deviceURL, err := d.client.URL(10)
 	if err != nil {
 		return models.Device{}, err
 	}
@@ -144,7 +145,7 @@ func (d *deviceRestClient) DeviceForName(name string, ctx context.Context) (mode
 }
 
 func (d *deviceRestClient) DevicesByLabel(label string, ctx context.Context) ([]models.Device, error) {
-	deviceURL, err := d.endpointer.URL(10)
+	deviceURL, err := d.client.URL(10)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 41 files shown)


michaelestrin:
recheck

clients/interfaces/builder.go
```diff
@@ -0,0 +1,26 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package interfaces
+
+// RestClientBuilder is the interface for types that need to define some way to retrieve state information about services.
+// This information can be anything that must be determined at runtime, whether it is unknown or simply not yet known.
+type RestClientBuilder interface {
```
michaelestrin: Name is ambiguous to me.  "builder" is a relatively generic term.  Could be mistaken for the GoF Builder pattern (which it most definitely isn't a specific implementation of).

Could this be simply a URL interface?

clients/metadata/device.go
```diff
@@ -17,11 +17,13 @@ package metadata
 import (
 	"context"
 	"encoding/json"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/rest"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
```
michaelestrin: Sort imports.

clients/metadata/device.go
```diff
@@ -74,35 +76,22 @@ type DeviceClient interface {
 }
 
 type deviceRestClient struct {
-	url      string
-	endpoint clients.Endpointer
+	client interfaces.RestClientBuilder
```
michaelestrin: Match field name to contract name; i.e. RestClientBuilder type -> restClientBuilder (or builder) field name.

Commented once; may apply elsewhere.

clients/rest/consul.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"errors"
+	"time"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+)
+
+type consulClient struct {
```
michaelestrin: Add docblock.

clients/rest/consul.go
```diff
@@ -0,0 +1,72 @@
...
+	if c.initialized {
+		return c.url, nil
+	}
+
+	timer := time.After(time.Duration(c.timeout) * time.Second)
+	ticker := time.Tick(500 * time.Millisecond)
+	for {
+		select {
+		case <-timer:
+			return "", errors.New("client not yet initialized")
```
michaelestrin: Export error text into public constant.  More accurate description might be "unable to initialize client in %c.timeout seconds"

clients/rest/consul.go
```diff
@@ -0,0 +1,72 @@
+/*******************************************************************************
```
michaelestrin: Create corresponding consul_test.go with unit tests.

clients/rest/factory.go
```diff
@@ -0,0 +1,30 @@
+/*******************************************************************************
```
michaelestrin: Create corresponding factory_test.go with unit tests.

clients/rest/local.go
```diff
@@ -0,0 +1,35 @@
+/*******************************************************************************
```
michaelestrin: Create corresponding local_test.go with unit tests.

clients/rest/local.go
```diff
@@ -0,0 +1,35 @@
...
+ * the License.
+ *******************************************************************************/
+
+package rest
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+)
+
+type localClient struct {
```
michaelestrin: Add docblock.

clients/metadata/device.go
```diff
@@ -20,8 +20,11 @@ import (
 	"net/url"
 	"strconv"
 
-	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/rest"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
```
michaelestrin: Resort imports.  All are go-mod-core-contracts-based.  Empty line should be removed.

clients/rest/factory_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+)
+
+func TestClientFactoryLocal(t *testing.T) {
+	actualClient := ClientFactory(types.EndpointParams{UseRegistry: false}, nil)
```
michaelestrin: In production usage, I believe ClientFactory's m parameter will always be non-nil.  Test should reflect that.

clients/rest/local_test.go
```diff
@@ -0,0 +1,45 @@
...
+	expectedURL := "http://brandonforster.com"
+	actualClient := newLocalClient(types.EndpointParams{Url: expectedURL})
+
+	if actualClient.url != expectedURL {
+		t.Fatalf("expected URL %s, found URL %s", expectedURL, actualClient.url)
+	}
+}
+
+func TestLocalClient_URLPrefix(t *testing.T) {
+	expectedURL := "http://brandonforster.com"
```
michaelestrin: Use `http://domain.com` instead.

Commented once; occurs elsewhere.

clients/rest/local_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+)
+
+func TestNewLocalClient(t *testing.T) {
+	expectedURL := "http://brandonforster.com"
+	actualClient := newLocalClient(types.EndpointParams{Url: expectedURL})
+
+	if actualClient.url != expectedURL {
+		t.Fatalf("expected URL %s, found URL %s", expectedURL, actualClient.url)
+	}
```
michaelestrin: Tests shouldn't reference internal implementation.  `url` is private.  

I think the purpose of this test is to ensure `newLocalClient` returns a non-nil `*localClient` value.  Your assertion should be updated to reflect that.

clients/rest/registry_test.go
```diff
@@ -0,0 +1,82 @@
...
+package rest
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+)
+
```
michaelestrin: Add a test that ensures `newRegistryClient` returns a non-nil `*registryClient` value.

brandonforster:
> The `rest` package at `clients/rest` isn't named well. It encapsulates a factory and different strategies for obtaining a client URL. `rest` doesn't describe this responsibility well.

Thanks. I've done my best to redo it according to https://blog.golang.org/package-names, let me know if you think there's still more to do there.

clients/urlclient/local_test.go
```diff
@@ -0,0 +1,52 @@
...
+	if actualClient == nil {
+		t.Fatal("nil returned from newLocalClient")
+	}
+
+	expectedType := reflect.TypeOf(&localClient{})
+	clientType := reflect.TypeOf(actualClient)
+
+	if clientType != expectedType {
+		t.Fatalf("expected type %T, found %T", expectedType, actualClient)
+	}
```
michaelestrin: Reflection isn't necessary since `newRegistryClient` returns *registryClient test will never fail.

clients/urlclient/registry_test.go
```diff
@@ -0,0 +1,98 @@
...
+	if actualClient == nil {
+		t.Fatal("nil returned from newRegistryClient")
+	}
+
+	expectedType := reflect.TypeOf(&registryClient{})
+	clientType := reflect.TypeOf(actualClient)
+
+	if clientType != expectedType {
+		t.Fatalf("expected type %T, found %T", expectedType, actualClient)
+	}
```
michaelestrin: Reflection isn't necessary since `newRegistryClient` returns *registryClient test will never fail.

clients/urlclient/local_test.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func TestLocalClient_URLPrefix(t *testing.T) {
+	expectedURL := "http://domain.com"
+	client := newLocalClient(types.EndpointParams{Url: expectedURL})
+
+	actualURL, err := client.Prefix()
```
michaelestrin: With rename of `localClient.URLPrefix()` to `localClient.Prefix`, `client` here should be renamed here to `urlClient` to match the changes made elsewhere (and more importantly, to provide context for what `Prefix()` actually returns).

michaelestrin:
Merged with https://github.com/edgexfoundry/go-mod-core-contracts/pull/197

Issue #196, opened 2020-01-08
Refactor client endpoints and fix race condition


brandonforster:
While working https://github.com/edgexfoundry/edgex-go/issues/2237 it was discovered that the race condition causing the URL to be empty exists for every endpoint on every client. Most have timing such that it won't manifest, but the underlying race condition must be addressed.

PR #197, opened 2020-01-09
Refactor Client Code, part 2 of several

brandonforster:
This PR is the second part of several working on #196.

This PR refactors all of metadata's clients and updates their unit tests.

To test, actuate the endpoints in the metadata device client. The blackbox tests for command should work for that.

### DIFF
--- a/clients/metadata/device_service.go
+++ b/clients/metadata/device_service.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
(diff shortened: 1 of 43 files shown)


clients/metadata/device_service.go
```diff
@@ -41,58 +39,60 @@ type DeviceServiceClient interface {
...
+	if err != nil {
+		return err
 	}
+
+	_, err = clients.PutRequest(serviceURL+"/"+id+"/lastconnected/"+strconv.FormatInt(time, 10), nil, ctx)
+	return err
 }
 
-// Helper method to request and decode a device service
-func (s *deviceServiceRestClient) requestDeviceService(url string, ctx context.Context) (models.DeviceService, error) {
```
brandonforster: I removed this helper method because it is only used once and is unlikely to be used anywhere else.

PR #198, opened 2020-01-16
Refactor Client Code, part 3/7: core-data

brandonforster:
This PR is the third part of several working on #196.

This PR refactors all of core-data's clients and updates their unit tests.

To test, actuate the endpoints in the core-data REST client.

### DIFF
--- a/clients/coredata/event.go
+++ b/clients/coredata/event.go
@@ -15,9 +15,7 @@
  *
  *******************************************************************************/
 
-/*
-Package coredata provides clients used for integration with the core-data service.
-*/
+// coredata provides clients used for integration with the core-data service.
 package coredata
 
 import (
@@ -29,6 +27,7 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/urlclient"
 	"github.com/edgexfoundry/go-mod-core-contracts/models"
 )
 
@@ -50,7 +49,7 @@ type EventClient interface {
 	EventsForDeviceAndValueDescriptor(deviceId string, vd string, limit int, ctx context.Context) ([]models.Event, error)
 	// Add will post a new event
 	Add(event *models.Event, ctx context.Context) (string, error)
-	//AddBytes posts a new event using an array of bytes, supporting encoding of the event by the caller.
+	// AddBytes posts a new event using an array of bytes, supporting encoding of the event by the caller.
 	AddBytes(event []byte, ctx context.Context) (string, error)
 	// DeleteForDevice will delete events by the specified device name
 	DeleteForDevice(id string, ctx context.Context) error
@@ -68,37 +67,24 @@ type EventClient interface {
 }
 
 type eventRestClient struct {
-	url      string
-	endpoint interfaces.Endpointer
+	urlClient interfaces.URLClient
 }
 
 // NewEventClient creates an instance of EventClient
 func NewEventClient(params types.EndpointParams, m interfaces.Endpointer) EventClient {
-	e := eventRestClient{endpoint: m}
-	e.init(params)
-	return &e
-}
-
-func (e *eventRestClient) init(params types.EndpointParams) {
-	if params.UseRegistry {
-		go func(ch chan string) {
-			for {
-				select {
-				case url := <-ch:
-					e.url = url
-				}
-			}
-		}(e.endpoint.Monitor(params))
-	} else {
-		e.url = params.Url
-	}
+	return &eventRestClient{urlClient: urlclient.New(params, m)}
 }
 
 // Helper method to request and decode an event slice
-func (e *eventRestClient) requestEventSlice(url string, ctx context.Context) ([]models.Event, error) {
-	data, err := clients.GetRequest(url, ctx)
+func (e *eventRestClient) requestEventSlice(urlSuffix string, ctx context.Context) ([]models.Event, error) {
+	urlPrefix, err := e.urlClient.Prefix()
 	if err != nil {
-		return []models.Event{}, err
+		return nil, err
+	}
+
+	data, err := clients.GetRequest(urlPrefix+urlSuffix, ctx)
+	if err != nil {
+		return nil, err
 	}
 
 	eSlice := make([]models.Event, 0)
@@ -107,8 +93,13 @@ func (e *eventRestClient) requestEventSlice(url string, ctx context.Context) ([]
 }
 
 // Helper method to request and decode an event
-func (e *eventRestClient) requestEvent(url string, ctx context.Context) (models.Event, error) {
-	data, err := clients.GetRequest(url, ctx)
+func (e *eventRestClient) requestEvent(urlSuffix string, ctx context.Context) (models.Event, error) {
+	urlPrefix, err := e.urlClient.Prefix()
+	if err != nil {
+		return models.Event{}, err
+	}
+
+	data, err := clients.GetRequest(urlPrefix+urlSuffix, ctx)
 	if err != nil {
 		return models.Event{}, err
 	}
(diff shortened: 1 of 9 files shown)


clients/coredata/event.go
```diff
@@ -68,37 +67,24 @@ type EventClient interface {
...
+func (e *eventRestClient) requestEventSlice(urlSuffix string, ctx context.Context) ([]models.Event, error) {
+	urlPrefix, err := e.urlClient.Prefix()
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := clients.GetRequest(urlPrefix+urlSuffix, ctx)
 	if err != nil {
-		return []models.Event{}, err
+		return nil, err
```
michaelestrin: Not sure if this should be nil or an empty slice of models.Event.

Golang idiom is to return nil.  EdgeX idiom is to return empty slice of models.Event.  We need to be consistent with EdgeX idiom.

See comments immediately above and below.

May occur elsewhere.

clients/coredata/event.go
```diff
@@ -68,37 +67,24 @@ type EventClient interface {
...
+	return &eventRestClient{urlClient: urlclient.New(params, m)}
 }
 
 // Helper method to request and decode an event slice
-func (e *eventRestClient) requestEventSlice(url string, ctx context.Context) ([]models.Event, error) {
-	data, err := clients.GetRequest(url, ctx)
+func (e *eventRestClient) requestEventSlice(urlSuffix string, ctx context.Context) ([]models.Event, error) {
+	urlPrefix, err := e.urlClient.Prefix()
+	if err != nil {
+		return nil, err
```
michaelestrin: See comment immediately below.

clients/coredata/event.go
```diff
@@ -68,37 +67,24 @@ type EventClient interface {
...
+	}
+
+	data, err := clients.GetRequest(urlPrefix+urlSuffix, ctx)
 	if err != nil {
-		return []models.Event{}, err
+		return nil, err
 	}
 
 	eSlice := make([]models.Event, 0)
```
michaelestrin: See comment immediately above. eSlice is returned regardless -- is an empty slice when err is not nil.

clients/coredata/event.go
```diff
@@ -119,65 +110,117 @@ func (e *eventRestClient) requestEvent(url string, ctx context.Context) (models.
...
-func (e *eventRestClient) EventsForDeviceAndValueDescriptor(deviceId string, vd string, limit int, ctx context.Context) ([]models.Event, error) {
-	return e.requestEventSlice(e.url+"/device/"+url.QueryEscape(deviceId)+"/valuedescriptor/"+url.QueryEscape(vd)+"/"+strconv.Itoa(limit), ctx)
+func (e *eventRestClient) EventsForDeviceAndValueDescriptor(
+	deviceId string,
+	vd string,
+	limit int,
+	ctx context.Context) ([]models.Event, error) {
+
+	return e.requestEventSlice(
+		"/device/"+url.QueryEscape(deviceId)+"/valuedescriptor/"+url.QueryEscape(vd)+"/"+strconv.Itoa(limit), ctx)
```
michaelestrin: Awkward folding.  Prefer

```
	return e.requestEventSlice(
		"/device/"+
			url.QueryEscape(deviceId)+
			"/valuedescriptor/"+
			url.QueryEscape(vd)+
			"/"+strconv.Itoa(limit),
		ctx,
	)
```

clients/coredata/event_test.go
```diff
@@ -23,12 +23,12 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
-	"time"
+
+	"github.com/ugorji/go/codec"
```
michaelestrin: Move to below github.com/edgexfoundry imports (and separate from them with a blank line).

clients/coredata/reading.go
```diff
@@ -109,49 +102,87 @@ func (r *readingRestClient) requestReading(url string, ctx context.Context) (mod
...
-func (r *readingRestClient) ReadingsForNameAndDevice(name string, deviceId string, limit int, ctx context.Context) ([]models.Reading, error) {
-	return r.requestReadingSlice(r.url+"/name/"+url.QueryEscape(name)+"/device/"+url.QueryEscape(deviceId)+"/"+strconv.Itoa(limit), ctx)
+func (r *readingRestClient) ReadingsForNameAndDevice(
+	name string,
+	deviceId string,
+	limit int,
+	ctx context.Context) ([]models.Reading, error) {
+
+	return r.requestReadingSlice(
+		"/name/"+url.QueryEscape(name)+"/device/"+url.QueryEscape(deviceId)+"/"+strconv.Itoa(limit), ctx)
```
michaelestrin: Awkward folding.  See earlier comment.  May occur elsewhere.

clients/urlclient/registry_test.go
```diff
@@ -55,11 +47,11 @@ func TestRegistryClient_URLPrefix(t *testing.T) {
 
 func TestRegistryClient_URLPrefixInitialized(t *testing.T) {
 	expectedURL := "http://domain.com"
-	client := newRegistryClient(types.EndpointParams{}, mockEndpoint{}, 100)
-	client.initialized = true
-	client.url = expectedURL
+	urlClient := newRegistryClient(types.EndpointParams{}, mockEndpoint{}, 100)
+	urlClient.initialized = true
+	urlClient.url = expectedURL
```
michaelestrin: I didn't catch these in your earlier PR.  Tests shouldn't query or mutate a receiver's internal state.  Is there a way you can write this test without doing so (and without adding additional test-specific methods to the receiver to facilitate the test)?

brandonforster:
recheck

clients/urlclient/registry_test.go
```diff
@@ -55,11 +47,15 @@ func TestRegistryClient_URLPrefix(t *testing.T) {
 
 func TestRegistryClient_URLPrefixInitialized(t *testing.T) {
 	expectedURL := "http://domain.com"
-	client := newRegistryClient(types.EndpointParams{}, mockEndpoint{}, 100)
-	client.initialized = true
-	client.url = expectedURL
+	urlClient := newRegistryClient(types.EndpointParams{}, mockEndpoint{}, 100)
+
+	// set up prerequisite condition
```
michaelestrin: Is comment accurate?

clients/urlclient/registry_test.go
```diff
@@ -55,11 +47,15 @@ func TestRegistryClient_URLPrefix(t *testing.T) {
...
+	actualURL, err := urlClient.Prefix()
+	if err != nil {
+		t.Fatalf("unexpected error in precondition %s", err.Error())
+	}
```
michaelestrin: This implementation duplicates the implementation that immediately follows it.

clients/urlclient/registry_test.go
brandonforster: Yes, we're trying to test what happens when the initialized flag is set. The only way to do that without modifying state is to call prefix twice.

clients/urlclient/registry_test.go
brandonforster: I'm trying to ensure that initialized url retrieval (which is a different code path) works as expected. Do you have any suggestions on how to better accomplish that without modifying the internal state?

michaelestrin: `TestRegistryClient_URLPrefix` may not actually test what you expect.  Behavior is indeterminate and dependent on timing.  In one execution, it may exercise the `c.Initialized` path in `Prefix()`.  In another, it may fall into the for loop.

I'd replace the interfaces.Endpointer mock with a stub that creates the channel returned by Monitor when the stub is constructed.  I'd add a `SendToChannel()` method to the stub that allows the test to control the timing of the monitor's population of the `registryClient`'s URL property.  I'd then use the new method in both `TestNewRegistryClient` and `TestRegistryClient_URLPrefix` to facilitate the desired tests.  I'd include comments within the tests that document the intent behind the specific testing implementations.

As an aside, the goroutine in `mockTimeoutEndpoint.Monitor()` is superfluous and should be removed.

brandonforster:
@michaelestrin let me know if this looks good and if so, I'll kick off the rebase cascade and update the names of the URL variables to urlPrefix.

clients/urlclient/registry_test.go
michaelestrin: To be handled in follow up issue https://github.com/edgexfoundry/go-mod-core-contracts/issues/204

michaelestrin:
PR content included in https://github.com/edgexfoundry/go-mod-core-contracts/pull/202.

PR #202, opened 2020-01-20
Refactor Client Code, part 7/7: scheduler

brandonforster:
This PR is the seventh part of several working on #196.

This PR refactors all of scheduler's client methods.

### DIFF
--- a/clients/scheduler/interval.go
+++ b/clients/scheduler/interval.go
@@ -12,9 +12,7 @@
  * the License.
  *******************************************************************************/
 
-/*
-Package scheduler provides clients used for integration with the support-scheduler service.
-*/
+// scheduler provides clients used for integration with the support-scheduler service.
 package scheduler
 
 import (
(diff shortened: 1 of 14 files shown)


clients/scheduler/interval.go
```diff
@@ -49,84 +46,100 @@ type IntervalClient interface {
...
-	return clients.UpdateRequest(s.url, interval, ctx)
+func (ic *intervalRestClient) Intervals(ctx context.Context) ([]models.Interval, error) {
+	return ic.requestIntervalSlice("", ctx)
 }
 
-//
-// Helper functions
-//
+func (ic *intervalRestClient) Update(interval models.Interval, ctx context.Context) error {
+	url, err := ic.urlClient.Prefix()
```
michaelestrin: Variable naming in prior PRs (and elsewhere in this same file) is `urlPrefix`; rename `url` here for consistency.  Applies to any/all calls to Prefix().

brandonforster: The distinction I made was that when there's no suffix added I just named it "url"

michaelestrin: I see.  That wasn't immediately apparent to me.  Still prefer consistent naming of variable returned by Prefix(); `urlPrefix` makes the most sense to me.

brandonforster:
Do not merge until #201 is merged.

PR #203, opened 2020-01-21
Add tests for interval client

brandonforster:
While working #196, it was noticed that the scheduler package has no unit tests whatsoever. This PR adds a basic suite of unit tests, taking it from 0% to ~75% coverage.

Dependent on #205.

### DIFF
--- a/clients/scheduler/interval_action_test.go
+++ b/clients/scheduler/interval_action_test.go
@@ -1,0 +1,335 @@
+/*******************************************************************************
+ * Copyright 2020 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package scheduler
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+var testIntervalAction1 = models.IntervalAction{
+	ID:         testID1,
+	Created:    123,
+	Modified:   123,
+	Origin:     123,
+	Name:       "testName",
+	Interval:   "123",
+	Parameters: "123",
+	Target:     "testNombre",
+	Protocol:   "123",
+	HTTPMethod: "get",
+	Address:    "localhost",
+	Port:       2700,
+	Path:       "123",
+	Publisher:  "123",
+	User:       "123",
+	Password:   "123",
+	Topic:      "123",
+}
+
+var testIntervalAction2 = models.IntervalAction{
+	ID:         testID2,
+	Created:    321,
+	Modified:   321,
+	Origin:     321,
+	Name:       "testNombre",
+	Interval:   "321",
+	Parameters: "321",
+	Target:     "testName",
+	Protocol:   "321",
+	HTTPMethod: "post",
+	Address:    "127.0.0.1",
+	Port:       3000,
+	Path:       "321",
+	Publisher:  "321",
+	User:       "321",
+	Password:   "321",
+	Topic:      "321",
+}
+
+func TestIntervalActionRestClient_Add(t *testing.T) {
+	ts := testHttpServer(t, http.MethodPost, clients.ApiIntervalActionRoute)
+
+	defer ts.Close()
+
+	url := ts.URL + clients.ApiIntervalActionRoute
+
+	params := types.EndpointParams{
+		ServiceKey:  clients.SupportSchedulerServiceKey,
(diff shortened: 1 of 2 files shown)


brandonforster:
> For complete coverage for `interval_action`, could we add tests which cover the error cases? For example there a calls to ` iac.urlClient.Prefix()` but cases where the `err != nil` is never tested:
> 
> https://github.com/edgexfoundry/go-mod-core-contracts/blob/6d3824e5f1709ea5a130ab5a8cda199335ae1838/clients/scheduler/interval_action.go#L107-L110

These calls are about to be eliminated in the next PR.

michaelestrin:
> I personally would be OK with not covering those error cases in this PR and creating a follow-up issue which targets testing those marshaling errors everywhere in this project in a consistent manner.

Let's spend the extra cycles to do it right the first time.

clients/scheduler/interval_action_test.go
```diff
@@ -0,0 +1,438 @@
...
+		}
+
+		data, err := json.Marshal(testIntervalAction1)
+		if err != nil {
+			t.Fatalf("marshaling error: %s", err.Error())
+		}
+		w.Write(data)
+	}))
+
+	badJSONServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
AnthonyMBonafide: I am not a fan of starting a server to just be able to unit test a client. One reason is that is actually requires the test to obtain a port to listen on which is can possibly introduce issues depending on the environment. I think this is showing that the client should be a little more flexible and updated for better test-ability.

During the last round of comments I was suggesting to allow a caller to inject an "unmarshaler" and I think this still applies to this round of review. This would allow for easy testing, since you can inject an implementation which would just return an error and during happy path tests you can inject `json.Unmarshal` or some wrapper making these tests a lot simpler. This gives the added benefit of the client now being able to support many more content types(Open–closed principle).

Here is an example:

Create a type alias which will allow the caller to inject the unmarshaling function they wish to be applied. For example:

```go
type ResponseUnmarshaler func(data []byte, v interface{}) error
```

Then in your constructor(s), you can have the caller pass in that type:
```go

type intervalActionRestClient struct {
	urlClient interfaces.URLClient
        unmarshaler ResponseUnmarshaler
}

// NewIntervalActionClient creates an instance of IntervalActionClient
func NewIntervalActionClient(params types.EndpointParams, m interfaces.Endpointer, unmarshaler ResponseUnmarshaler) IntervalActionClient {
	return &intervalActionRestClient{urlClient: urlclient.New(params, m), unmarshaler: unmarshaler}
}
```

To support backwards compatibility you can keep the existing constructor and have it use `json.Unmarshal` as a default. This would work since all existing users of the construct are using it for `JSON` anyway or else it would not work.

Now, instead of creating a running server just to cover test cases where serialization has failed you can inject a custom `ResponseUnmarshaler` which always returns an `error`:
```go
// mockUnmarshalerError returns an error when unmarshaling is attempted.
func mockUnmarshalerError([]byte, interface{}) error {
	return errors.New("unmarshal error")
}

NewIntervalActionClient(params, endpointer, mockUnmarshalerError)
```

and the methods for `intervalActionRestClient` would refer to `intervalActionRestClient.unmarshaller` rather than `json.Unmarshal`:

```go
// Helper method to request and decode an interval action
func (iac *intervalActionRestClient) requestIntervalAction(
	urlSuffix string,
	ctx context.Context) (models.IntervalAction, error) {

	data, err := clients.GetRequest(urlSuffix, ctx, iac.urlClient)
	if err != nil {
		return models.IntervalAction{}, err
	}

	ia := models.IntervalAction{}
	err = ia.unmarshaler(data, &ia)
	if err != nil {
		return models.IntervalAction{}, err
	}

	return ia, nil
}
```

Furthermore, you can reuse the `ResponseUnmarshaler` type other packages which have similar structure to this allowing consistency in the project, so I think a good place for this is probably `request.go` to be used by all clients. All of this also applies to marshaling as well. An example of other clients having similar structure is:
https://github.com/edgexfoundry/go-mod-core-contracts/blob/4b1aba370a2a780425ec0b989b3dc7884f6a1289/clients/coredata/event.go#L102 


For more examples you can refer to the work in `go-mod-messaging` here is a good starting point. https://github.com/AnthonyMBonafide/go-mod-messaging/blob/a775bafe5b2419b22a509946a8df0703ec825d1e/internal/pkg/mqtt/client.go#L43-L69

michaelestrin: > I am not a fan of starting a server to just be able to unit test a client.

I concur.

> During the last round of comments I was suggesting to allow a caller to inject an "unmarshaler" and I think this still applies to this round of review.

Not sure why the original suggested direction wasn't incorporated in this latest round of changes.

brandonforster:
#208 created to address issues raised in this PR.

In my repo, this branch: 

https://github.com/brandonforster/go-mod-core-contracts/tree/unmarshaler_tests

shows what a half step of just doing the unmarshaler would look like. I can submit that code for review, but I would prefer submitting this code that takes the "wrong" (but consistent with the rest of the repo) approach today, and doing it right everywhere tomorrow.

clients/scheduler/interval_action_test.go
```diff
@@ -0,0 +1,438 @@
...
+		expectedURL := clients.ApiIntervalActionRoute + "/" + testID1
+		if r.URL.EscapedPath() != expectedURL {
+			t.Fatalf("expected uri path is %s, actual uri path is %s", expectedURL, r.URL.EscapedPath())
+		}
+
+		data, err := json.Marshal(testIntervalAction1)
+		if err != nil {
+			t.Fatalf("marshaling error: %s", err.Error())
+		}
+		w.Write(data)
```
AnthonyMBonafide: Sorry, I missed this before. Can we explicitly ignore the returned values?

```go
_, _ = w.Write(data)
```

This applies for the other tests which use this method.

PR #205, opened 2020-01-23
Centralize URLClient logic to the Request package

### LINKED ISSUE #196: Refactor client endpoints and fix race condition

brandonforster:
Fixes #196.

This is the final PR for #196. This PR moves all calls of `URLPrefix` to the `request` package to reduce duplication.

There are two places I would like to call out for special review:

`ValueDescriptorsUsage()` in metadata/provision_watcher.go and
`sendLog()` in `logger/logger.go.

These two functions required me creating a new URLClient that returns an empty string to fulfill the contract and preserve functionality. I did my best, but am not tremendously happy with this solution and would like any suggestions on how to improve it.

### DIFF
--- a/clients/command/client.go
+++ b/clients/command/client.go
@@ -48,8 +48,7 @@ func NewCommandClient(params types.EndpointParams, m interfaces.Endpointer) Comm
 }
 
 func (cc *commandRestClient) Get(deviceId string, commandId string, ctx context.Context) (string, error) {
-	body, err := clients.GetRequest("/"+deviceId+"/command/"+commandId, ctx, cc.urlClient)
-	return string(body), err
+	return cc.getRequestJSONBody("/"+deviceId+"/command/"+commandId, ctx)
 }
 
 func (cc *commandRestClient) Put(deviceId string, commandId string, body string, ctx context.Context) (string, error) {
@@ -61,12 +60,7 @@ func (cc *commandRestClient) GetDeviceCommandByNames(
 	commandName string,
 	ctx context.Context) (string, error) {
 
-	body, err := clients.GetRequest("/name/"+deviceName+"/command/"+commandName, ctx, cc.urlClient)
-	if err != nil {
-		return "", err
-	}
-
-	return string(body), nil
+	return cc.getRequestJSONBody("/name/"+deviceName+"/command/"+commandName, ctx)
 }
 
 func (cc *commandRestClient) PutDeviceCommandByNames(
@@ -76,4 +70,10 @@ func (cc *commandRestClient) PutDeviceCommandByNames(
 	ctx context.Context) (string, error) {
 
 	return clients.PutRequest("/name/"+deviceName+"/command/"+commandName, []byte(body), ctx, cc.urlClient)
+}
+
+func (cc *commandRestClient) getRequestJSONBody(urlSuffix string, ctx context.Context) (string, error) {
+	body, err := clients.GetRequest(urlSuffix, ctx, cc.urlClient)
+
+	return string(body), err
 }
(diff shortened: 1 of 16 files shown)


clients/command/client.go
```diff
@@ -48,36 +48,25 @@ func NewCommandClient(params types.EndpointParams, m interfaces.Endpointer) Comm
...
 
-	url, err := cc.urlClient.Prefix()
+	body, err := clients.GetRequest("/name/"+deviceName+"/command/"+commandName, ctx, cc.urlClient)
 	if err != nil {
 		return "", err
 	}
 
-	body, err := clients.GetRequest(url+"/name/"+deviceName+"/command/"+commandName, ctx)
-	return string(body), err
+	return string(body), nil
```
michaelestrin: Inconsistent with `Get` implementation above.

Look at creating a `clients.GetRequestReturnString` (not the best name, think about it and come up with something better) that does the cast for you.  This would further simplify code in lots of place from something like what you have above to:

```
return clients.GetRequestReturnString("/name/"+deviceName+"/command/"+commandName, ctx, cc.urlClient)
```

clients/coredata/event.go
```diff
@@ -98,12 +93,7 @@ func (e *eventRestClient) requestEventSlice(urlSuffix string, ctx context.Contex
 
 // Helper method to request and decode an event
 func (e *eventRestClient) requestEvent(urlSuffix string, ctx context.Context) (models.Event, error) {
```
michaelestrin: Please create a new issue to refactor context.Context usage to make it idiomatic (i.e. `ctx context.Context` as first parameter in a method/function parameter list).  Note that this will break the module's API contracts.  Reference this comment in the new issue.

clients/coredata/value_descriptor.go
```diff
@@ -161,7 +151,15 @@ func (v *valueDescriptorRestClient) ValueDescriptorsUsage(names []string, ctx co
...
-	data, err := clients.GetRequest(u.String(), ctx)
+
+	// create a new URL client with an empty URL to fulfil the GetRequest contract while relying wholly on
+	// our parsed URL for the actual endpoint data
+	emptyURLClient := urlclient.New(types.EndpointParams{
+		UseRegistry: false,
+		Url:         "",
+	}, nil)
+
+	data, err := clients.GetRequest(u.String(), ctx, emptyURLClient)
```
michaelestrin: Don't do this.

It looks like you're basically after `GetRequest`'s functionality without the `urlClient.Prefix()` call (and related error checking).  Look at `clients.GetRequest` to extract functionality that you're after into a separate function there.  Call that new function here.  It's a much cleaner and clearer approach.

clients/logger/logger.go
```diff
@@ -249,8 +251,15 @@ func (lc edgeXLogger) buildLogEntry(logLevel string, msg string, args ...interfa
 
 // Send the log as an http request
 func (lc edgeXLogger) sendLog(logEntry models.LogEntry) {
+	// create a new URL client with an empty URL to fulfil the PostJsonRequest contract while relying wholly on
+	// logTarget for the actual endpoint data
+	emptyURLClient := urlclient.New(types.EndpointParams{
+		UseRegistry: false,
+		Url:         "",
+	}, nil)
+
```
michaelestrin: Don't do this.  See my comment on this PR that deals with similar behavior implemented elsewhere.

Issue #206, opened 2020-01-24
Refactor usage of Context in entire repo to make it idiomatic


brandonforster:
This repo does not use Golang best practices and does not have `ctx context.Context` as the first method/function parameter when it is used. This should be done throughout the repo.

This will be a breaking change.

Discussion and discovery from @michaelestrin :
https://github.com/edgexfoundry/go-mod-core-contracts/pull/205/files/4b1aba370a2a780425ec0b989b3dc7884f6a1289#diff-5b56cccfe7a55a3fe3b288bdb434df3e

Related: https://github.com/edgexfoundry/go-mod-core-contracts/pull/205

PR #207, opened 2020-01-29
Implement strategy pattern on RegistryClient URLClient

### LINKED ISSUE #204: Implement strategy pattern on retry algorithm in registry implementation of URLClient
While working #196, it was decided that the retry algorithm should be customizable and configurable for different use cases. This issue will implement a new interface and contain three strategies for the algorithm:

1. the current retry algorithm
2. a "No Retry" algorithm that fails fast and does not attempt to retry
3. a null implementation intended for clients that know they will use a URLClient implementation that does not use the retry algorithm. This implementation should panic on all calls to it.

Additionally, the unit tests for the registry client should be rewritten to be deterministic and rely on new methods written for the strategies to ensure that the strategy is executed when expected.

brandonforster:
Fixes #204.

Includes deterministic tests.

To verify, run `make test`.

Signed-off-by: Brandon Forster <me@brandonforster.com>

### DIFF
--- a/clients/urlclient/factory.go
+++ b/clients/urlclient/factory.go
@@ -19,12 +19,13 @@ package urlclient
 import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/urlclient/retry"
 )
 
 // New provides the correct concrete implementation of the URLClient given the params provided.
 func New(params types.EndpointParams, m interfaces.Endpointer) interfaces.URLClient {
 	if params.UseRegistry {
-		return newRegistryClient(params, m, 10)
+		return newRegistryClient(params, m, retry.NewPeriodicRetry(500, 10))
 	}
 	return newLocalClient(params)
 }
--- a/clients/urlclient/factory.go
+++ b/clients/urlclient/factory.go
@@ -23,9 +23,13 @@ import (
 )
 
 // New provides the correct concrete implementation of the URLClient given the params provided.
-func New(params types.EndpointParams, m interfaces.Endpointer) interfaces.URLClient {
-	if params.UseRegistry {
-		return newRegistryClient(params, m, retry.NewPeriodicRetry(500, 10))
+func New(
+	endpointParams types.EndpointParams,
+	m interfaces.Endpointer,
+	urlClientParams types.URLClientParams) interfaces.URLClient {
+
+	if endpointParams.UseRegistry {
+		return newRegistryClient(endpointParams, m, retry.NewPeriodicRetry(urlClientParams))
 	}
-	return newLocalClient(params)
+	return newLocalClient(endpointParams)
 }
--- a/clients/urlclient/interfaces/retry.go
+++ b/clients/urlclient/interfaces/retry.go
@@ -22,7 +22,7 @@ package interfaces
 // the URL that should be passed back to the RegistryClient should be returned by the RetryStrategy.
 type RetryStrategy interface {
 	// Retry defines the actual behavior of the RegistryClient and how it processes the URL.
-	Retry(isInitialized *bool, url *string) (string, error)
+	Retry(url *string) (string, error)
 
 	// IsLocked communicates whether the value of the URL is currently being updated or not.
 	IsLocked() bool
--- a/clients/urlclient/interfaces/retry.go
+++ b/clients/urlclient/interfaces/retry.go
@@ -1,0 +1,32 @@
+/*******************************************************************************
+ * Copyright 2020 Dell Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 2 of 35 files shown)


clients/urlclient/factory.go
```diff
@@ -19,12 +19,13 @@ package urlclient
...
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/urlclient/retry"
 )
 
 // New provides the correct concrete implementation of the URLClient given the params provided.
 func New(params types.EndpointParams, m interfaces.Endpointer) interfaces.URLClient {
 	if params.UseRegistry {
-		return newRegistryClient(params, m, 10)
+		return newRegistryClient(params, m, retry.NewPeriodicRetry(500, 10))
```
brandonforster: Upon submitting this I remembered a tertiary goal of this issue was to get rid of these magic numbers. Working on a commit now.

brandonforster:
With the second commit included this will be a breaking change in edgex-go.

clients/urlclient/interfaces/retry.go
```diff
@@ -0,0 +1,32 @@
...
+// the URL that should be passed back to the RegistryClient should be returned by the RetryStrategy.
+type RetryStrategy interface {
+	// Retry defines the actual behavior of the RegistryClient and how it processes the URL.
+	Retry(url *string) (string, error)
+
+	// IsLocked communicates whether the value of the URL is currently being updated or not.
+	IsLocked() bool
+
+	// SetLock updates the value of the lock.
+	SetLock(value bool)
```
AnthonyMBonafide: Are these methods needed for the interface? I see it:

1. A caller shouldn't need invoke either of these methods, but rather only invoke `Retry`
1. The functionality provided are move useful for the implementation to have guards internally, if applicable.
1. Not all implementations will use this. For example, **I think** the `once` implementation could simply be:
```go
// Retry is designed to poll for the data one time. If successful, it returns the data, otherwise it returns an error.
func (o *once) Retry(url *string) (string, error) {
	if url != "" {
		return *url, nil
	}

	return "", errors.NewTimeoutError()
}
```

I bring this up because even though this pattern is useful, the caller still has the burden of managing the locks outside of the `RetryStrategy`:
```go
// newRegistryClient returns a pointer to a registryClient.
// A pointer is used so that when using configuration from a registry, the Prefix can be updated asynchronously.
func newRegistryClient(
	params types.EndpointParams,
	m interfaces.Endpointer,
	strategy urlClientInterfaces.RetryStrategy) *registryClient {

	e := registryClient{
		initialized: false,
		strategy:    strategy,
	}

	go func(ch chan string) {
		for {
			select {
			case url := <-ch:
				e.url = url
				strategy.SetLock(false)
			}
		}
	}(m.Monitor(params))

	return &e
}
``` 

I wonder if we do decide to reduce this down to one method, can this be achieved with type aliasing and higher-order functions. For example:
```go
type RetryStrategy func(*string) error
```

And providing your implementations `once` and `periodic` via higher-order functions which create the `RetryStrategy`. For example:
```go
type RetryStrategy func(*string) error

func NewOnceRetryStrategy() RetryStrategy{
	return func(value *string) error{
		if value != nil {
			return nil
		}
		
		return errors.NewTimeoutError()
	}
}

func NewPeriodicRetryStrategy(timeout time.Duration, ticker time.Duration) RetryStrategy{
	return func(value *string) error{
		if value != nil{
			return nil
		}
		
		// Do waiting logic and return if present
		
		// Else return error
		return errors.NewTimeoutError()
	}	
}
```

This is just a thought, I am not suggesting to do this, just bringing it up as an alternative implementation. I hope all of this makes sense and I have the correct understanding of the implementation.

clients/urlclient/interfaces/retry.go
```diff
@@ -0,0 +1,32 @@
...
+package interfaces
+
+// RetryStrategy defines some way to verify that a RegistryClient has received data.
+// It accomplishes this by monitoring the values pointed to in its parameters,
+// which communicate the state of the RegistryClient.
+// When the values pointed to by the parameters are in the desired state,
+// the URL that should be passed back to the RegistryClient should be returned by the RetryStrategy.
+type RetryStrategy interface {
+	// Retry defines the actual behavior of the RegistryClient and how it processes the URL.
+	Retry(url *string) (string, error)
```
AnthonyMBonafide: Can we change the name from `url` to something more general like `value`. I only suggest that because some of the comments use the word `poll` and that mixed with `url` can possibly lead to believe that the `RetryStrategy` is polling that URL. I made this mistake when I jumped right into the code at first. If this is silly then ignore this comment.

brandonforster:
Thanks for your review @AnthonyMBonafide. @michaelestrin and I decided that there's some more work that should be done here, and that will be captured in a PR addressing #204, #206, and #209. I'll probably do #208 since a lot of code will be touched, but no promises.

In the interim, this PR is incomplete w/r/t direction we're trying to go, so I'm going to close it without merge.

PR #211, opened 2020-02-03
Remove Endpointer construct. Add RetryStrategy to RegistryClient.

### LINKED ISSUE #204: Implement strategy pattern on retry algorithm in registry implementation of URLClient

### LINKED ISSUE #206: Refactor usage of Context in entire repo to make it idiomatic

### LINKED ISSUE #209: Refactor client code to reduce coupling on Endpointer
Currently, there is very tight coupling with the Endpointer construct, and the Endpointer is hardly used. It is standing in the way of architectural goals and should be eliminated to streamline the contract set for clients.

This will be a breaking change in edgex-go.

brandonforster:
Fixes #204 
Fixes #206 
Fixes #209 

This PR will be a breaking change in `edgex-go`.


clients/interfaces/urlstream.go
```diff
@@ -12,28 +12,6 @@
...
-
-func TestClientFactoryRegistry(t *testing.T) {
-	actualClient := New(types.EndpointParams{UseRegistry: true}, mockEndpoint{})
-	_, isRegistryClient := actualClient.(*registryClient)
-
-	if !isRegistryClient {
-		t.Fatalf("expected type %T, found %T", registryClient{}, actualClient)
-	}
-}
+type URLStream string
```
brandonforster: [I structured this this way because of this SO answer](https://stackoverflow.com/questions/36820663/type-alias-for-channel-in-go-works-strange)

clients/urlclient/retry/once.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/urlclient/errors"
+)
+
+// once is designed to poll for the data one time. If successful, it returns the data, otherwise it returns an error.
+type once struct {
+	isLocked bool
+}
+
+func NewOnce() *once {
```
michaelestrin: If you put this into its own package and give the package a descriptive name, you should be able to rename this to `New` and have it make sense.  Further, your receiver struct name and references could also be genericized.

clients/urlclient/retry/once.go
```diff
@@ -0,0 +1,50 @@
...
+// Retry is designed to poll for the data one time. If successful, it returns the data, otherwise it returns an error.
+func (o *once) Retry(url *string) (string, error) {
+	if !o.isLocked {
+		return *url, nil
+	}
+
+	return "", errors.NewTimeoutError()
+}
+
+// IsLocked communicates whether the value of the URL is currently being updated or not.
```
michaelestrin: IsLocked is awkward; can you think of a better name?  It's really capturing the concept of whether or not the url has been initialized yet.

clients/logger/logger.go
```diff
@@ -251,7 +251,7 @@ func (lc edgeXLogger) buildLogEntry(logLevel string, msg string, args ...interfa
 // Send the log as an http request
 func (lc edgeXLogger) sendLog(logEntry models.LogEntry) {
 	go func() {
-		_, err := clients.PostJsonRequestWithURL(lc.logTarget, logEntry, context.Background())
+		_, err := clients.PostJsonRequestWithURL(context.Background(), lc.logTarget, logEntry)
```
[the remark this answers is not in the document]
brandonforster: Context is currently being used not for thread control or any higher purpose than a simple data store. We can trivially eliminate most of these data store use cases (such as `ContentType` header storage and retrieval) but its usage as a place to put the `correlation` header is a trickier knot to untangle. I'm going to investigate what removing it there looks like, and if I'm successful and it's not a huge change set I'll include it in this work. If it's a big change, or if it's not clear that it's possible to untangle the correlation header, I'll file an issue and we can take a look at it later.

clients/urlclient/interfaces/retry.go
```diff
@@ -0,0 +1,32 @@
...
+
+// interfaces defines contracts specific to the urlclient package
+package interfaces
+
+// RetryStrategy defines some way to verify that a RegistryClient has received data.
+// It accomplishes this by monitoring the values pointed to in its parameters,
+// which communicate the state of the RegistryClient.
+// When the values pointed to by the parameters are in the desired state,
+// the URL that should be passed back to the RegistryClient should be returned by the RetryStrategy.
+type RetryStrategy interface {
```
AnthonyMBonafide: I do not know if my previous comments  still apply: https://github.com/edgexfoundry/go-mod-core-contracts/pull/207#discussion_r373183961

brandonforster: I think it's a neat idea, but it puts so much onus on the consumer to generate implementation that I don't want to do it. I've been playing with this idea today and I couldn't find an iteration I was happy with that balanced leaking implementation (what we're doing, tbh, with the lock state stuff that exists to aid testing) and something like your proposal that swings back to the other extreme.

clients/metadata/device.go
```diff
@@ -22,69 +22,69 @@ import (
...
 // DeviceClient defines the interface for interactions with the Device endpoint on core-metadata.
 type DeviceClient interface {
 	// Add creates a new device
-	Add(dev *models.Device, ctx context.Context) (string, error)
+	Add(ctx context.Context, dev *models.Device) (string, error)
 	// Delete eliminates a device for the specified ID
-	Delete(id string, ctx context.Context) error
+	Delete(ctx context.Context, id string) error
 	// DeleteByName eliminates a device for the specified name
 	DeleteByName(name string, ctx context.Context) error
```
michaelestrin: context.Context should be first parameter.

Please review PR and correct all instances.

brandonforster:
Rebased.

models/interval.go
```diff
@@ -32,8 +32,8 @@ type Interval struct {
 	Timestamps  Timestamps
 	ID          string `json:"id,omitempty"`
 	Name        string `json:"name,omitempty"`      // non-database identifier for a shcedule (*must be quitue)
-	Start       string `json:"start,omitempty"`     // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
-	End         string `json:"end,omitempty"`       // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
+	Start       string `json:"start,omitempty"`     // Start time i ISO 8601 format YYYYMMDD'T'HHmmss 	@JSONFormat(shape = JSONFormat.Shape.STRING, pattern = "yyyymmdd'T'HHmmss")
```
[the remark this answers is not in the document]
brandonforster: They're in comments and aren't run. We can safely get rid of them.

After a quick Google, it looks like they're format strings for the Jackson library in Java Spring. In comments. In Go.

clients/urlclient/retry/periodic/periodic.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"time"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/urlclient/errors"
+)
+
+// periodic is designed to poll for the data on a regular frequency until the timeout happens.
+// tickTime defines the interval in milliseconds for how often the URL should be queried.
+// timeout defines the interval in seconds for how long the algorithm should query before giving up.
+type periodic struct {
```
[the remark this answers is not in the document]
michaelestrin: You missed receiver variable name. ;)

clients/urlclient/retry/once/once.go
```diff
@@ -0,0 +1,48 @@
...
+
+// once is designed to poll for the data one time. If successful, it returns the data, otherwise it returns an error.
+package once
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/urlclient/errors"
+)
+
+// once is designed to poll for the data one time. If successful, it returns the data, otherwise it returns an error.
+type once struct {
```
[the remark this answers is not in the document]
michaelestrin: You missed receiver variable name. ;)

clients/urlclient/interfaces/retry.go
michaelestrin: Agree this could be better structured.  Going to dismiss your review so we can get this code merged.

PR #215, opened 2020-02-13
Added mock required in EdgeX Go.

brandonforster:
Added mock that was deleted in #211 that is required by edgex-go.

Signed-off-by: Brandon Forster <me@brandonforster.com>


brandonforster:
Closed. This mock will be moved to where it is used in EdgeX Go.

PR #223, opened 2020-03-04
Add ValueType, MediaType, FloatEncoding to Reading

### LINKED ISSUE #193: Current model of Readings does not provide Datatype of Value
The [Reading structure ](https://github.com/edgexfoundry/go-mod-core-contracts/blob/892c76594632f1888adc7f8949698420825669b9/models/reading.go#L22) is missing a field for Datatype of Sensor Value. Presently all the Values are converted to String format for each Reading. But the **datatype** of value **is not retained**, and I **cannot** find any way to **recover the datatype of Value**.

Proposed Solution:
A new field can be introduced in the structure to retain the datatype.  For this purpose, we can reuse the [ValueType](https://github.com/edgexfoundry/device-sdk-go/blob/0749bf2bbbaf424e6ee3aee44608466ddf380fa6/pkg/models/commandvalue.go#L24) enum from Device SDK Go.

PS: Let me know if there is a way to recover datatype or the thought-process behind not retaining the Value's datatype in the first place.

*** Solution [discussed ](https://github.com/edgexfoundry/go-mod-core-contracts/issues/193#issuecomment-593481163)in Device Services WG 2-Mar-2020 ***

jdharms:
Fixes #193.  Add some fields to Reading value as discussed
on 3/2/20 DS Working Group call and documented in discussion
at https://github.com/edgexfoundry/go-mod-core-contracts/issues/193

## PR Checklist

- [x] Tests for the changes have been updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The Reading structure is missing a field for Datatype of Sensor Value.

Issue Number:
https://github.com/edgexfoundry/go-mod-core-contracts/issues/193

## What is the new behavior?
Reading type has fields for ValueType, FloatEncoding, and MediaType.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/models/reading.go
+++ b/models/reading.go
@@ -20,31 +20,37 @@ import (
 
 // Reading contains data that was gathered from a device.
 type Reading struct {
-	Id          string `json:"id,omitempty" codec:"id,omitempty"`
-	Pushed      int64  `json:"pushed,omitempty" codec:"pushed,omitempty"`   // When the data was pushed out of EdgeX (0 - not pushed yet)
-	Created     int64  `json:"created,omitempty" codec:"created,omitempty"` // When the reading was created
-	Origin      int64  `json:"origin,omitempty" codec:"origin,omitempty"`
-	Modified    int64  `json:"modified,omitempty" codec:"modified,omitempty"`
-	Device      string `json:"device,omitempty" codec:"device,omitempty"`
-	Name        string `json:"name,omitempty" codec:"name,omitempty"`
-	Value       string `json:"value,omitempty"  codec:"value,omitempty"`            // Device sensor data value
-	BinaryValue []byte `json:"binaryValue,omitempty" codec:"binaryValue,omitempty"` // Binary data payload
-	isValidated bool   // internal member used for validation check
+	Id            string `json:"id,omitempty" codec:"id,omitempty"`
+	Pushed        int64  `json:"pushed,omitempty" codec:"pushed,omitempty"`   // When the data was pushed out of EdgeX (0 - not pushed yet)
+	Created       int64  `json:"created,omitempty" codec:"created,omitempty"` // When the reading was created
+	Origin        int64  `json:"origin,omitempty" codec:"origin,omitempty"`
+	Modified      int64  `json:"modified,omitempty" codec:"modified,omitempty"`
+	Device        string `json:"device,omitempty" codec:"device,omitempty"`
+	Name          string `json:"name,omitempty" codec:"name,omitempty"`
+	Value         string `json:"value,omitempty" codec:"value,omitempty"` // Device sensor data value
+	ValueType     string `json:"valueType,omitempty" codec:"valueType,omitempty"`
+	FloatEncoding string `json:"floatEncoding,omitempty" codec:"floatEncoding,omitempty"`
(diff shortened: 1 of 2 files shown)


models/reading.go
```diff
@@ -20,31 +20,37 @@ import (
...
+		Created       int64   `json:"created"`
+		Origin        int64   `json:"origin"`
+		Modified      int64   `json:"modified"`
+		Device        *string `json:"device"`
+		Name          *string `json:"name"`
+		Value         *string `json:"value"`
+		ValueType     *string `json:"valueType"`
+		FloatEncoding *string `json:"floatEncoding"`
+		BinaryValue   []byte  `json:"binaryValue"`
+		MediaType     *string `json:"mediaType"`
```
AnthonyMBonafide: As per Trevor's comment: https://github.com/edgexfoundry/go-mod-core-contracts/issues/193#issuecomment-593481163

Will we need to add some additional logic to the `Validate` method to ensure the data is correct?

jdharms: I think you’re referring to the media type only being filled in for binary data? I think that’s not something we’re looking to enforce necessarily, but I’ll look into it and get back to you.

tsconn23: Per Anthony's comment, it's a good idea to add logic to `Validate()` so that 
- if `ValueType == Binary` then `MediaType` should not be empty.
- if `ValueType == Float` then `FloatEncoding` should not be empty.

AnthonyMBonafide: Should we also enforce that only the proper "content" is populated? For example,

- if ValueType == Binary then FloatEncoding should be empty.
- if ValueType == Float then MediaType should be empty.

This would protect the user from providing incorrect data, and also protect us from wasting storage space on useless data.

jdharms: @tsconn23 How should I do the comparisons on the ValueType property? Should I define string constants in this file for clients of core-contracts to use to fill it in?  Something like 
```
// Constants related to Reading ValueTypes
const (
	ValueTypeBool    = "Bool"
	ValueTypeString  = "String"
	ValueTypeUint8   = "Uint8"
	ValueTypeUint16  = "Uint16"
	ValueTypeUint32  = "Uint32"
	ValueTypeUint64  = "Uint64"
	ValueTypeInt8    = "Int8"
	ValueTypeInt16   = "Int16"
	ValueTypeInt32   = "Int32"
	ValueTypeInt64   = "Int64"
	ValueTypeFloat32 = "Float32"
	ValueTypeFloat64 = "Float64"
	ValueTypeBinary  = "Binary"
)
```
?

jdharms: @tsconn23 Hm, alright.  I only ask because it would seem we need some way to check the values inside ValueType if we're going to perform the checks such as `if ValueType == Float then FloatEncoding should not be empty.`

For the binary one I suppose we can just check to see if the BinaryData value is non-empty, but I don't see any way around inspecting the value for the Float case.  Unless someone has a better idea I suppose I'll create the constants I suggested above.

The enum referenced above is an integer enum--won't be useful here.

jdharms: @AnthonyMBonafide @tsconn23 Implemented these checks and added test cases.  Please take a look and give any feedback you may have!

Issue #226, opened 2020-03-29
ProvisionWatcher embeds DeviceProfile and DeviceService


tonyespy:
The `ProvistionWatcher` model embeds both a `DeviceProfile` and `DeviceService` (as does a Device). These should be changed to strings, and should just contain the names of both objects, as there's no reason the device SDKs need the entire models when performing discovery.

**Note** - I looked, but couldn't find the related issue to address this in `Device`. We probably need to wait for Hanio to make the change for `Device`, as it would cause a breaking API change. The same can't really be said for `ProvisionWatcher`s as they haven't been used for anything until this release (Geneva).

jpwhitemn:
Per Tony's comment on #826, is this Geneva or Hanoi?

tonyespy:
@jpwhitemn As we haven't yet implemented V2 watchers this issue really only applies to 1.x, so my vote would be to close this out, as this would be a breaking change to V1.

@cloudxxx8 we should just ensure that when we do implement provision watchers for V2, we make sure to only include device profile and device service names vs. embedding the objects.

cloudxxx8:
@tonyespy sure, that's the plan.
> 
> @cloudxxx8 we should just ensure that when we do implement provision watchers for V2, we make sure to only include device profile and device service names vs. embedding the objects.

cloudxxx8:
V2 design has addressed this issue, please see
https://github.com/edgexfoundry/edgex-go/pull/2982
https://github.com/edgexfoundry/go-mod-core-contracts/issues/371

PR #230, opened 2020-03-31
Replace Profile and Service in ProvisionWatcher

cloudxxx8:
Replace Profile field with ProfileName and change the type to
string.
Replace Service field with DeviceServiceName and change the type to
string.
fix https://github.com/edgexfoundry/go-mod-core-contracts/issues/226

Signed-off-by: Cloud Tsai <cloudxxx8@gmail.com>

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Simplify the data model and reduce the size

Issue Number:
https://github.com/edgexfoundry/go-mod-core-contracts/issues/226

- [x] Yes
- [ ] No

## Other information

### DIFF
--- a/clients/metadata/provision_watcher_test.go
+++ b/clients/metadata/provision_watcher_test.go
@@ -29,11 +29,11 @@ import (
 // Test adding a provision watcher using the urlClient
 func TestAddProvisionWatcher(t *testing.T) {
 	se := models.ProvisionWatcher{
-		Id:             "1234",
-		Name:           "Test name for provision watcher",
-		Profile:        models.DeviceProfile{},
-		Service:        models.DeviceService{},
-		OperatingState: models.Enabled,
+		Id:                "1234",
+		Name:              "Test name for provision watcher",
+		ProfileName:       "Test name for device profile",
+		DeviceServiceName: "Test name for device profile",
+		OperatingState:    models.Enabled,
 	}
 
 	addingProvisionWatcherID := se.Id
--- a/models/provisionwatcher.go
+++ b/models/provisionwatcher.go
@@ -16,7 +16,6 @@ package models
 
 import (
 	"encoding/json"
-	"reflect"
 )
 
 type ProvisionWatcher struct {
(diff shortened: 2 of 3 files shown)


tsconn23:
@brandonforster Having worked on the ProvisionWatcher operations in Core, what is the effect on the proposed changes in this PR? Is anything used in the current ProvisionWatcher from the Device Profile/Service other than the name?

cloudxxx8:
we need some more work on device-sdk-go to apply this change if we think it should be in Geneva.
I believe it also affects the existing implementation in core-metadata.  we should decide whether we should spend time to reduce this object transportation size.

iain-anderson:
Would it be better to leave the model object as-is, and instead define a DTO in core which replaces profile and service with structs containing "name" strings. Then map the results from eg dbClient.GetProvisionWatcherById() into the DTO and return that from the rest operation?

brandonforster:
@tsconn23 It appears as though the Mongo driver does not make any usage of either of these fields anywhere in the code except to do transforms between the DB model and the contract model.

The Redis driver, however, makes extensive use of both in the ways that you'd expect for basic CRUD. I'd say that's a product of flawed Mongo implementation (that I am discovering numerous other problems with in the course of working #2464 and intend to address either as part of that or as follow on) rather than an intentional architectural decision. I'd vote for something like @iain-anderson's solution, or simply pushing this off until Hanoi and make this breaking change then.

However, I'd also add that I'm basically reading tea leaves here, and it's tremendously hard to say what conscious decisions about how this should work *were actually* made.

tsconn23:
Discussed during Device Service WG 6-Apr-2020 and based on current workload prior to code freeze, decision was to defer this PR. Tagged as `Hanoi` for visibility of inclusion in V2 API work.

cloudxxx8:
Thanks, @tsconn23 
Close this PR, and I will open a new one for V2 API work

Issue #235, opened 2020-04-10
Create core data DTOs and data models for V2 API


lindseysimple:
Based on the discussion in April 9th Core Working Group Meeting, the DTOs and data models of V2 API specification will be implemented in the `go-mod-core-contracts` modules.

The core data DTOs and models addressed in this issue are designed based on the V2 API swagger definition
[core-data.yaml](https://github.com/edgexfoundry/edgex-go/blob/master/api/openapi/v2/core-data.yaml)

Change includes:

- Event object
- Reading object

Originally issue: https://github.com/edgexfoundry/edgex-go/issues/2475

PR #236, opened 2020-04-10
Create core data DTOs and models for V2 API

lindseysimple:
Create core data DTOs and models for Event & Reading.
Fix: #235

Signed-off-by: Lindsey Cheng <beckysocute@gmail.com>

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #235 

- [ ] Yes
- [x] No

#### validator package:
Package [go-playground/validator](https://github.com/go-playground/validator) is used to implement value validations for structs and individual fields based on tags.
Reasons for use this package:
- It can also handle Cross-Field and Cross-Struct validation for nested structs and has the ability to dive into arrays and maps of any type.
- It has 5.4k stars on github
- It has 141 releases and the latest release [v10.3.0](https://github.com/go-playground/validator/releases/tag/v10.3.0) just came out in May, 2020.
- Currently the validation functions are defined in each individual contract. By using the `validator` package, the effort to define/maintain the validation function for each individual contract can be reduced. We can use the built-in validation features in the package to validate each field, and only need to define the validation condition using the `validate` tag in the struct definition.

---------------------------------------------------------------------------------
The initial PR was raised in https://github.com/edgexfoundry/edgex-go/pull/2476.

Based on the discussion in April 9th Core Working Group Meeting, the DTOs and new models for V2 API specification should be defined in the `go-mod-core-contracts`.


lindseysimple:
Update data model and DTO design for `Reading`.

lindseysimple:
Changes updated:
- add `Event`  model `UnmarshalJSON` function
- add `AddEventRequest` DTO validation function
- add Unit test function:`TestEvent_UnmarshalJSON` and `TestAddEventRequest_Validate`

tsconn23:
@lindseysimple Please note that the verify build is broken. If you'd like someone to look at this PR, please correct the build issue.

lindseysimple:
Build issue has been fixed. Thanks @tsconn23 for the reminder.

jpwhitemn:
putting this on hold until we get Geneva out.  Then anyone can lift the hold flag.

go.mod
```diff
@@ -4,9 +4,14 @@ require (
 	github.com/fxamacker/cbor/v2 v2.2.0
 	github.com/go-kit/kit v0.8.0
 	github.com/go-logfmt/logfmt v0.4.0 // indirect
+	github.com/go-playground/locales v0.13.0
+	github.com/go-playground/universal-translator v0.17.0
 	github.com/go-stack/stack v1.8.0 // indirect
 	github.com/google/uuid v1.1.0
+	github.com/leodido/go-urn v1.1.0 // indirect
```
tsconn23, line 11: @lindseysimple Be advised this dependency itself has a dependency on a project that is [no longer maintained](https://github.com/pmezard/go-difflib). Per recent discussions in the architecture meetings, this probably won't be acceptable.

https://github.com/leodido/go-urn/blob/a0f5013415294bb94553821ace21a1a74c0298cc/go.sum#L3

Are there alternatives?

go.mod
```diff
@@ -4,9 +4,14 @@ require (
...
 	github.com/stretchr/testify v1.3.0
+	gopkg.in/go-playground/assert.v1 v1.2.1 // indirect
```
tsconn23, line 13: There is apparently a v2 version of this library found [here](https://github.com/go-playground/assert). Given the `//indirect` comment another dependency probably pulled this in. Maybe the one below.

go.mod
```diff
@@ -4,9 +4,14 @@ require (
...
+	gopkg.in/go-playground/validator.v9 v9.31.0
```
tsconn23, line 14: The current version of this library is [v10.2.0](https://github.com/go-playground/validator/releases/tag/v10.2.0). Recommend seeing if that works. Looking at the `go.sum` you're still going to have a problem though with the transitive dependency to [pmezard/go-difflib](https://github.com/pmezard/go-difflib) above.

v2/dtos/coredata/event.go
```diff
@@ -0,0 +1,117 @@
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package coredata
```
tsconn23, line 6: My suggestion is to not include a service name in the package path. Instead, I would separate requests and responses agnostic of service. 
```
v2/dtos/requests
v2/dtos/responses
```

We have been discussing changing the name of  the CoreData service to SupportData which would then require us to change all of these package references, if we could do it at all since it would most likely be considered backwards incompatible.

v2/models/coredata/event.go
```diff
@@ -0,0 +1,20 @@
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package coredata
```
tsconn23, line 6: Same comment w/r/t inclusion of service name

v2/dtos/coredata/event.go
lindseysimple: Updated as suggested, thanks.

v2/models/coredata/event.go
lindseysimple: Updated as suggested.

go.mod
lindseysimple: [leodido/go-urn](github.com/leodido/go-urn) is updated to v1.2.0 after [go-playground/validator](https://github.com/go-playground/validator) is updated to v10.2.0. This line has been removed after running `go mod tidy`.

go.mod
lindseysimple: [go-playground/assert v2.0.1](gopkg.in/go-playground/assert) is used in [go-playground/validator](https://github.com/go-playground/validator). This line has been removed after running `go mod tidy`.

go.mod
lindseysimple: Updated [go-playground/validator](https://github.com/go-playground/validator) to v10.2.0 and it's still using the [pmezard/go-difflib](https://github.com/pmezard/go-difflib) module as you pointed out. 
Did some investigation but found no other good alternatives for the validator feature.

cloudxxx8: mezard/go-difflib is a small and simple module, and it is an indirect dependency.  go-playground/validator should adjust it if there is any trouble happening in the future.  We can discuss in the WG meeting to see whether we can accept it.

v2/dtos/responses/event.go
```diff
@@ -0,0 +1,51 @@
...
+// EventCountResponse object in the APIv2 specification.
+type EventCountResponse struct {
+	dtoBase.Response `json:",inline"`
+	Count            int
+	DeviceID         string `json:"deviceId"` // ID uniquely identifies a device
+}
+
+// EventResponse defines the Response Content for GET event DTOs.  This object and its properties correspond to the
+// EventResponse object in the APIv2 specification.
+type EventResponse struct {
```
cloudxxx8, line 41: is there any helper function convert from Event Model to Event DTO, as well as Reading?

v2/dtos/requests/event.go
```diff
@@ -0,0 +1,76 @@
...
+	a.Readings = addEvent.Readings
+
+	// validate AddEventRequest DTO
+	if err := a.Validate(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func ToEventModels(addRequests []AddEventRequest) (events []models.Event) {
```
cloudxxx8, line 63: We need to set up a naming convention.  Event can't be updated, but we will have more data models can be updated, such as Device.
Thus, I suggest to use `AddReqToEventModels`
We will have something like `AddReqtToDeviceModels `and `UpdateReqtToDeviceModels` in the future.

v2/dtos/requests/reading.go
```diff
@@ -0,0 +1,62 @@
...
+	Id            string   `json:"id,omitempty" codec:"id,omitempty"`
+	Pushed        int64    `json:"pushed,omitempty" codec:"pushed,omitempty"`   // When the data was pushed out of EdgeX (0 - not pushed yet)
+	Created       int64    `json:"created,omitempty" codec:"created,omitempty"` // When the reading was created
+	Origin        int64    `json:"origin,omitempty" codec:"origin,omitempty"`
+	Modified      int64    `json:"modified,omitempty" codec:"modified,omitempty"`
+	Device        string   `json:"device,omitempty" codec:"device,omitempty"`
+	Name          string   `json:"name,omitempty" codec:"name,omitempty"`
+	Labels        []string `json:"labels,omitempty" codec:"labels,omitempty"` // Custom labels assigned to a reading, added in the APIv2 specification.
+	BinaryReading `json:",inline"`
+	SimpleReading `json:",inline"`
```
cloudxxx8, line 22: we need validation in the Reading DTO as well.

v2/dtos/requests/event_test.go
```diff
@@ -0,0 +1,112 @@
...
+		Device:   v2.TestDeviceName,
+		Origin:   v2.TestOriginTime,
+		Readings: []models.Reading{},
+	}}
+	tests := []struct {
+		name        string
+		addEvents   []AddEventRequest
+		expectError bool
+	}{
+		{"valid AddEventRequest", valid, false},
```
lenny-intel, line 101: Added error case(s)

v2/dtos/requests/event_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			readingModel := ToEventModels(tt.addEvents)
+			if !reflect.DeepEqual(expectedEventModel, readingModel) {
```
lenny-intel, line 106: Use testify assert.Equal

v2/dtos/requests/reading.go
```diff
@@ -0,0 +1,62 @@
...
+type BinaryReading struct {
+	BinaryValue []byte `json:"binaryValue,omitempty" codec:"binaryValue,omitempty"` // Binary data payload
+	MediaType   string `json:"mediaType,omitempty" codec:"mediaType,omitempty"`     // indicates what the content type of the binaryValue property is
+}
+
+// An event reading for a simple data type
+// SimpleReading object in the APIv2 specification.
+type SimpleReading struct {
+	Value         string `json:"value,omitempty" codec:"value,omitempty"`                 // Device sensor data value
+	ValueType     string `json:"valueType,omitempty" codec:"valueType,omitempty"`         // Indicates the datatype of the value property
```
lenny-intel, line 36: ValueType is also set when reading is Binary, correct? So maybe it should be part of BaseReading?

v2/dtos/requests/reading.go
```diff
@@ -0,0 +1,62 @@
...
+	FloatEncoding string `json:"floatEncoding,omitempty" codec:"floatEncoding,omitempty"` // Indicates how a float value is encoded
+}
+
+// convert Reading DTO to ReadingInterface struct model
```
lenny-intel, line 40: ReadingInterface ?? `ReadingInterface struct model` => `Reading model`

v2/dtos/requests/reading.go
```diff
@@ -0,0 +1,62 @@
...
+func toReadingModel(r BaseReading, device string) models.Reading {
+	var readingModel models.Reading
+	br := models.BaseReading{
+		Device: device,
+		Name:   r.Name,
+		Labels: r.Labels,
+	}
+	if r.ValueType != "" {
```
lenny-intel, line 48: Can't ValueType be `"Binary"`??

v2/dtos/requests/reading_test.go
```diff
@@ -0,0 +1,52 @@
...
+		name        string
+		reading     BaseReading
+		expectError bool
+	}{
+		{"valid Event", valid, false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			readingModel := toReadingModel(tt.reading, v2.TestDeviceName)
+			if !reflect.DeepEqual(expected, readingModel) {
```
lenny-intel, line 46: use assert.Equal()

v2/dtos/requests/reading_test.go
```diff
@@ -0,0 +1,52 @@
...
+			Device: v2.TestDeviceName,
+			Name:   v2.TestReadingName,
+		},
+		Value:     v2.TestValue,
+		ValueType: v2.TestValueType,
+	}
+	tests := []struct {
+		name        string
+		reading     BaseReading
+		expectError bool
```
lenny-intel, line 39: Remove, no error ever returned.

v2/dtos/requests/reading_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/v2"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+)
+
+var testReading = BaseReading{
```
lenny-intel, line 17: change to `testSimpleReading` and add `testBinaryReading`

v2/dtos/requests/reading_test.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+		Value:     v2.TestValue,
+		ValueType: v2.TestValueType,
+	}
+	tests := []struct {
+		name        string
+		reading     BaseReading
+		expectError bool
+	}{
+		{"valid Event", valid, false},
```
lenny-intel, line 41: Add test for binary.

v2/dtos/responses/event.go
```diff
@@ -0,0 +1,51 @@
...
+	dtoBase.Response `json:",inline"`
+	Event            Event
+}
+
+// UpdateEventPushedByChecksumResponse defines the Response Content for PUT event as pushed DTO. This object and its properties correspond to the
+// UpdateEventPushedByChecksumResponse object in the APIv2 specification.
+type UpdateEventPushedByChecksumResponse struct {
+	dtoBase.Response `json:",inline"`
+	Checksum         string `json:"checksum"`
```
lenny-intel, line 50: Why do we need to return the `Checksum` that we sent in the request?

v2/models/reading.go
```diff
@@ -0,0 +1,47 @@
...
+	BinaryValue []byte `json:"binaryValue,omitempty" codec:"binaryValue,omitempty"` // Binary data payload
+	MediaType   string `json:"mediaType,omitempty" codec:"mediaType,omitempty"`     // indicates what the content type of the binaryValue property is
+}
+
+// An event reading for a simple data type
+// SimpleReading object in the APIv2 specification.
+type SimpleReading struct {
+	BaseReading   `json:",inline"`
+	Value         string `json:"value,omitempty" codec:"value,omitempty"`                 // Device sensor data value
+	ValueType     string `json:"valueType,omitempty" codec:"valueType,omitempty"`         // Indicates the datatype of the value property
```
lenny-intel, line 36: Same question here.. ValueType should be part of BaseReading since it can be set to "Binary"

v2/validator.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/go-playground/locales/en"
+	ut "github.com/go-playground/universal-translator"
+
+	"github.com/go-playground/validator/v10"
+)
+
+var (
+	val   *validator.Validate
+	trans ut.Translator
```
lenny-intel, line 17: Don't use global variables

v2/validator.go
```diff
@@ -0,0 +1,45 @@
...
+	_ = val.RegisterTranslation("required", trans, func(ut ut.Translator) error {
+		return ut.Add("required", "{0} cannot be blank", true) // see universal-translator for details
+	}, func(ut ut.Translator, fe validator.FieldError) string {
+		t, _ := ut.T("required", fe.Field())
+		return t
+	})
+	return val
+}
+
+func Validate(a interface{}) error {
```
lenny-intel, line 34: Make `val` a receiver

v2/validator.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/go-playground/validator/v10"
+)
+
+var (
+	val   *validator.Validate
+	trans ut.Translator
+)
+
+func NewValidator() *validator.Validate {
```
lenny-intel, line 20: What is this used for? It is very cryptic.

v2/validator.go
```diff
@@ -0,0 +1,45 @@
...
+		return ut.Add("required", "{0} cannot be blank", true) // see universal-translator for details
+	}, func(ut ut.Translator, fe validator.FieldError) string {
+		t, _ := ut.T("required", fe.Field())
+		return t
+	})
+	return val
+}
+
+func Validate(a interface{}) error {
+	err := val.Struct(a)
```
lenny-intel, line 35: Is this doing validation? if so, what is it validating?

v2/dtos/requests/event.go
```diff
@@ -0,0 +1,76 @@
...
+// UpdateEventPushedByChecksumRequest object in the APIv2 specification.
+type UpdateEventPushedByChecksumRequest struct {
+	dtoBase.Request `json:",inline"`
+	Checksum        string `json:"checksum"`
+}
+
+// Validate satisfies the Validator interface
+func (a AddEventRequest) Validate() error {
+	v2.NewValidator()
+	err := v2.Validate(a)
```
lenny-intel, line 35: What fields are being validated?

v2/validator.go
cloudxxx8: @lenny-intel we are using the new validation library.  the validation rules define in the Struct comment
For example,
```
type Request struct {
	CorrelationID string `json:"correlationId" validate:"required,uuid"`
	RequestID     string `json:"requestId" validate:"required,uuid"`
}
```

cloudxxx8: https://godoc.org/github.com/go-playground/validator

v2/dtos/requests/event.go
```diff
@@ -0,0 +1,78 @@
...
+	dtoBase "github.com/edgexfoundry/go-mod-core-contracts/v2/dtos/common/base"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+)
+
+// AddEventRequest defines the Request Content for POST event DTO. This object and its properties correspond to the
+// AddEventRequest object in the APIv2 specification.
+type AddEventRequest struct {
+	dtoBase.Request `json:",inline"`
+	Device          string               `json:"device" codec:"device,omitempty" validate:"required"`
+	Origin          int64                `json:"origin,omitempty" codec:"origin,omitempty" validate:"gte=1"`
```
tonyespy, line 22: How can this have a **omitempty** annotations and a **validate: gte=1**?

v2/dtos/requests/event.go
```diff
@@ -0,0 +1,78 @@
...
+	Readings        []common.BaseReading `json:"readings,omitempty" codec:"readings,omitempty"`
```
tonyespy, line 23: How can this ever be empty?

v2/dtos/responses/event.go
```diff
@@ -0,0 +1,51 @@
...
+// AddEventResponse defines the Response Content for POST event DTOs.  This object and its properties correspond to the
+// AddEventResponse object in the APIv2 specification.
+type AddEventResponse struct {
+	dtoBase.Response `json:",inline"`
+	ID               string `json:"id"` // ID uniquely identifies an event, for example a UUID
+}
+
+// Event represents a single measurable event read from a device
+type Event struct {
+	ID       string               `json:"id,omitempty" codec:"id,omitempty"`             // ID uniquely identifies an event, for example a UUID
```
tonyespy, line 22: Other the `Modified`, how can any of these be empty?

v2/models/reading.go
tonyespy: +1

v2/models/event.go
```diff
@@ -0,0 +1,20 @@
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package models
+
+// Event represents a single measurable event read from a device
+type Event struct {
```
tonyespy, line 9: Why are we adding new v2 versions of our internal models? This doesn't look like it's changed from the existing model?

v2/models/event.go
```diff
@@ -0,0 +1,20 @@
...
+	CorrelationId string
+	Checksum      string
+	ID            string    `json:"id,omitempty" codec:"id,omitempty"`             // ID uniquely identifies an event, for example a UUID
```
tonyespy, line 12: As these are internal models, why do they need to be annotated?

v2/validator.go
tonyespy: And it's public, please add godoc...

v2/models/reading.go
```diff
@@ -0,0 +1,47 @@
...
+	FloatEncoding string `json:"floatEncoding,omitempty" codec:"floatEncoding,omitempty"` // Indicates how a float value is encoded
+}
+
+// a abstract interface to be implemented by BinaryReading/SimpleReading
```
tonyespy, line 40: This needs a bit more explanation, it's not apparently obvious what this is being used for...

lenny-intel: +1. Took me a bit to realize what it is accomplishing. Is there a link to a pattern that was used to come up with this that could be included in the code?

go.mod
```diff
@@ -4,6 +4,9 @@ require (
 	github.com/fxamacker/cbor/v2 v2.2.0
 	github.com/go-kit/kit v0.8.0
 	github.com/go-logfmt/logfmt v0.4.0 // indirect
+	github.com/go-playground/locales v0.13.0
+	github.com/go-playground/universal-translator v0.17.0
```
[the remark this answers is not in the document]
lenny-intel: +1. We should discuss this in Core WG. @tsconn23

go.mod
```diff
@@ -4,6 +4,9 @@ require (
...
+	github.com/go-playground/validator/v10 v10.3.0
```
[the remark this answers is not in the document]
lenny-intel: +1. @tsconn23, Another discussion for Core WG.

v2/dtos/requests/event.go
lenny-intel: Ok, got it.

v2/dtos/responses/event.go
lindseysimple: Remove all the unnecessary `omitempty` tags except for `Modified` and `Pushed` as suggested.

v2/dtos/responses/event.go
lindseysimple: This function will be added in the later PRs for GET API, this PR addresses for `POST` Event.

v2/models/event.go
lindseysimple: As in the community core working group discussion we'll define separating models for v2 API to avoid confusion.

v2/models/event.go
lindseysimple: Removed the `json` and `codec` tags for models.

v2/models/reading.go
lindseysimple: As in the Open API doc https://github.com/edgexfoundry/edgex-go/blob/525c6541360a858effa5ffc827886954f889b9d1/api/openapi/v2/core-data.yaml#L114-L116, another `mediaType` field is defined in `BinaryReading` struct.

v2/models/reading.go
lindseysimple: We have done some research and found this way is probably more suitable in Golang to implement the inheritance (for `BinaryReading` and `SimpleReading` to inheriting the common `Reading` interface), no link or document for this kind of declaration. Any suggestion on how to explain?

lindseysimple: @lenny-intel `default` is a keyword in Golang (https://golang.org/ref/spec#Keywords). Would you please advise any other function name which is appropriate in this case? Thanks.

v2/dtos/requests/event.go
```diff
@@ -0,0 +1,79 @@
...
+	a.Readings = addEvent.Readings
+
+	// validate AddEventRequest DTO
+	if err := a.Validate(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Internal: AddEventReqToEventModels transforms the AddEventRequest DTO array to the Event model array
```
lenny-intel, line 65: This function is Public, not internal, as defined. If it should be internal then change name to `addEventReqToEventModels` or remove `Internal:` from the description.

v2/dtos/common/metrics.go
```diff
@@ -0,0 +1,27 @@
...
+type MetricsRequest struct {
+	RequestID string `json:"requestId" validate:"required,uuid"`
+}
+
+// MetricsResponse defines the providing memory and cpu utilization stats of the service.
+// This object and its properties correspond to the MetricsResponse object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/MetricsResponse
+type MetricsResponse struct {
+	BaseResponse   `json:",inline"`
+	MemAlloc       int  `json:"memAlloc" validate:"required"`
```
lenny-intel, line 20: Do we really need the validation annotation on the response DTO's?

v2/dtos/common/ping.go
```diff
@@ -0,0 +1,21 @@
...
+type PingRequest struct {
+	RequestID string `json:"requestId" validate:"required,uuid"`
+}
+
+// PingResponse defines the content of response content for POST Ping DTO
+// This object and its properties correspond to the Ping object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/PingResponse
+type PingResponse struct {
+	BaseResponse `json:",inline"`
+	Timestamp    string `json:"timestamp"`
```
lenny-intel, line 20: From recent testing it would be nice if ping returned the ServiceKey. I know this is a change to the API spec, but I think it is more useful than `Timestamp`. @tonyespy, @jpwhitemn  thoughts?

v2/models/reading.go
lenny-intel: Darn.. :-(
How about something like `defaultFunc` or `doNothing` ? Also add comment for the func stating that it's only purpose is for implementing the interface. THX!

cloudxxx8:
@lenny-intel @tonyespy @jpwhitemn 
After discussion with @lindseysimple , we decide to extract the logic to make a lightweight validation package on our own.  It would support the basic validation annotation, such as required, uuid, and larger than a certain number.  It could be achieved by the basic golang library without adding any new third party dependencies.  It might take a few days.  Do you agree with this approach?

go.mod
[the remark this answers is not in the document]
lindseysimple: Hi @tsconn23 @lenny-intel  , 
The default validation error message from the `validator` package is something like:
```
Key: 'AddEventRequest.BaseRequest.RequestID' Error:Field validation for 'RequestID' failed on the 'uuid' tag
Key: 'AddEventRequest.Device' Error:Field validation for 'RequestID' failed on the 'required' tag
```
Which only describes the error field and tag names and not quite representative. We'll try to make a general error handling function to translate the error messages to more understandable messages based on different validators as you suggested.

ccing @cloudxxx8

lindseysimple:
@lenny-intel @tonyespy 
The `en_translations` package is removed from go.mod and the [validator.go](https://github.com/edgexfoundry/go-mod-core-contracts/blob/1b58bf3054d7daa62a8799464b5cd267255e3d4a/v2/validator.go) file has been updated by adding a new `getErrorMessage` function to generate the customized error validation message.
Could you please review the change again? Thanks a lot for your help.

v2/validator.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Validate function will use the validator package to validate the struct annotation
+func Validate(a interface{}) error {
+	err := val.Struct(a)
+	// translate all error at once
+	if err != nil {
+		errs := err.(validator.ValidationErrors)
+		for _, e := range errs {
+			return getErrorMessage(e.Tag(), e.Field(), e.Param())
```
lenny-intel, line 26: If there are multiple validation errors, this will only return a message for the first error. Instead of returning, should build a a comma separated list of the errors and return the combined error message after the loop.

v2/validator.go
```diff
@@ -0,0 +1,45 @@
...
+func getErrorMessage(tag string, fieldName string, fieldValue string) error {
+	var msg string
+	switch tag {
+	case "uuid":
+		msg = fmt.Sprintf("%s field needs a uuid.", fieldName)
+	case "required":
+		msg = fmt.Sprintf("%s field is required.", fieldName)
+	case "len":
+		msg = fmt.Sprintf("The length of %s field is not %s.", fieldName, fieldValue)
+
```
lenny-intel, line 42: Need a default message that is built from the passed in error.

v2/validator.go
lenny-intel: Also, just pass `e` rather than the fields in `e` so the missing switch default case can depend on `e` for error message.

v2/validator.go
```diff
@@ -0,0 +1,52 @@
...
+func Validate(a interface{}) error {
+	err := val.Struct(a)
+	// translate all error at once
+	if err != nil {
+		errs := err.(validator.ValidationErrors)
+		var errMsg []string
+		for _, e := range errs {
+			errMsg = append(errMsg, getErrorMessage(e))
+		}
+		return NewErrContractInvalid(strings.Join(errMsg, ", "))
```
cloudxxx8, line 30: @lenny-intel may we use `\n` instead of comma to separate the error messages?  It would limit the error message cannot contain comma in the future.

lenny-intel: @cloudxxx8 , not wild about `\n` as it will impact how the message is logged. How about `; ` ?

v2/validator.go
```diff
@@ -0,0 +1,52 @@
...
+	var msg string
+	switch tag {
+	case "uuid":
+		msg = fmt.Sprintf("%s field needs a uuid", fieldName)
+	case "required":
+		msg = fmt.Sprintf("%s field is required", fieldName)
+	case "len":
+		msg = fmt.Sprintf("The length of %s field is not %s", fieldName, fieldValue)
+	default:
+		msg = fmt.Sprintf("%s field validation failed on the %s tag", fieldName, tag)
```
lenny-intel, line 49: Does the FieldError`e` have a built in error message that explains why validation failed? i.e. the default error messages we are overriding for `uuid`, `required` & `len`.  Should use it for the default case and then add additional overrides if needed when we determine the default message for a `tag` (not overridden) doesn't suffice?

v2/dtos/common/version.go
```diff
@@ -0,0 +1,21 @@
...
+// This object and its properties correspond to the VersionRequest object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/VersionRequest
+type VersionRequest struct {
+	RequestID string `json:"requestId" validate:"required,uuid"`
+}
+
+// VersionResponse defines the latest version supported by the service.
+// This object and its properties correspond to the VersionResponse object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/VersionResponse
+type VersionResponse struct {
```
lenny-intel, line 18: Can you please add the VersionSdkResponse that extends this response DTO with 
```
SdkVersion      string `json:"sdk_version"`
```

Both Device and App SDks will need this. Thx
@rsdmike FYI.

v2/dtos/common/metrics.go
```diff
@@ -0,0 +1,27 @@
...
+type MetricsRequest struct {
+	RequestID string `json:"requestId" validate:"required,uuid"`
+}
+
+// MetricsResponse defines the providing memory and cpu utilization stats of the service.
+// This object and its properties correspond to the MetricsResponse object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/MetricsResponse
+type MetricsResponse struct {
+	BaseResponse   `json:",inline"`
+	MemAlloc       int  `json:"memAlloc"`
```
rsdmike, line 20: shouldnt these be uint64?

v2/dtos/common/config.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+// ConfigRequest defines the request associated with the /config endpoint.
+// This object and its properties correspond to the ConfigRequest object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/ConfigRequest
+type ConfigRequest struct {
+	RequestID string `json:"requestId" validate:"required,uuid"`
```
rsdmike, line 12: For all of these requests that don't differ from BaseRequest -- do we need to create a separate struct?

v2/dtos/common/config.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// ConfigResponse defines the configuration for the targeted service.
+// This object and its properties correspond to the ConfigResponse object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/ConfigResponse
+type ConfigResponse struct {
+	BaseResponse `json:",inline"`
+	Config       string `json:"config"`
```
rsdmike, line 20: would this be better as an `interface{}` ?

v2/validator.go
lindseysimple: @lenny-intel Hi Lenny, I don't think there's a default error message in FieldError, the fields are describe in this [doc](https://godoc.org/github.com/go-playground/validator#FieldError) and only contains the information needed to generate the error message.

v2/dtos/common/config.go
lindseysimple: Have the same thought here, however, write the dto definitions based on the OpenAPI doc https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#, there are separate definitions which might be enhanced in the future.

v2/dtos/common/config.go
lindseysimple: @rsdmike Would you please explain which part should be an `interface{}` and the reason for it? Thanks

v2/dtos/common/metrics.go
lindseysimple: Updated `MemAlloc` and other fields as well, thanks.

v2/validator.go
cloudxxx8: @lenny-intel sounds good to me.  @lindseysimple let's use `;`

v2/validator.go
lenny-intel: Ok, thanks for looking into it.

v2/dtos/common/version.go
```diff
@@ -0,0 +1,28 @@
...
+	BaseResponse `json:",inline"`
+	Version      string `json:"version"`
+}
+
+// VersionSdkResponse defines the latest sdk version supported by the service.
+// This object and its properties correspond to the VersionSdkResponse object in the APIv2 specification:
+type VersionSdkResponse struct {
+	BaseResponse `json:",inline"`
```
lenny-intel, line 26: This needs to be `VersionResponse`, (instead of `BaseResponse`) since the App Services and Device Services (via SDKs) return both `Version` and `SdkVersion`

v2/dtos/common/config.go
lenny-intel: @lindseysimple , Each services conifg struct is different. Making this an `interface{}` allows us to populate the field with any services config struct and avoid the `embedded JSON`

rsdmike:
```
{
     "statusCode":200,
     "config": "{\"property\":\"field\"}" //an embedded json version of the config -- not so pretty
}
```
vs
```
{
     "statusCode":200,
     "config": {"property": "field"}   //properly a part of the json
}
```

lindseysimple: @lenny-intel @rsdmike , thanks for the explanation. Have updated `Config` to `Interface{}`. 
Would you please review again? Thanks.

v2/dtos/common/config.go
lenny-intel: @lindseysimple, @cloudxxx8  we are going to discuss the appropriateness of these request DTOs forcing normal GET requests to be PUT request in Thursday morning's Core WG.

rsdmike: These are POSTs but really they should probably be GETs - @jpwhitemn will talk with the team about this.

tonyespy:
OK, so I've done some more investigation into [go-difflib](https://github.com/pmezard/go-difflib) after our conversation at today's Core WG meeting.

First, it turns out that while we removed the **locale** and **universal-translator** packages from `go.mod` in this PR, they're both used by the **validator** package, so the end result is still the same. The usage of [go-playground/validator](https://github.com/go-playground/validator) results in an additional 11 imports being added to **go-core-mod-contracts**. I don't think we've yet walked through all of the new imports in this list and vetted them. For the purpose of this comment, I'm going to focus solely on **go-difflib** as it includes the following line at the top of it's [README.md](https://github.com/pmezard/go-difflib/blob/master/README.md) file:

> THIS PACKAGE IS NO LONGER MAINTAINED.

The project also has only a single release, and the last commit was two years ago!

So, I decided to look at how it gets pulled in by the validator package. It turns out validator doesn't directly import go-difflib, but it does import a package called go-urn which in turn imports the package testify (which provides a set of testing tools). go-urn is a URN validator package, and is only used if the JSON being validated is annotated with the tag "usn_2141". After reviewing go-urn, it's only use of the testify assert package is in _test.go files.

It also turns out that we use testify in our _test.go files in go-core-mod-contracts, which means we're already pulling go-difflib, but it's only being used for unit tests, none of our production code uses it.

So, long story short, I don't think we should block this PR due to go-difflib, but I do think we should ensure we've reviewed the rest of the new imports being introduced (see the new proposed [PDR](https://wiki.edgexfoundry.org/pages/viewpage.action?pageId=46760301)) for reviewingnew imports).

One last thought, perhaps when adding new imports which are only used for unit testing, we should add a comment to the corresponding `go.mod` which indicates this?

lindseysimple:
@lenny-intel @tonyespy @rsdmike, We've updated the change to remove the `ConfigRequest`, `MetricRequest`, 
 `PingRequest` and `VersionRequest` DTOs as those common API methods will be changed to `GET` instead of `POST`. 

Would you please merge this PR if no other concern?  Another PR will be opened in edgex-go to address the swagger file and core-data POST event changes after this PR being merged. Thanks.

cloudxxx8:
@tonyespy @tsconn23 you made change request in this PR, and we have addressed all of them.  please help review again when you have time.

jpwhitemn:
diff module/lib is approved via arch meeting.

lindseysimple:
Submit changes to rebase from `master` and rename the `ID` field to `Id` in `v2/models/event.go` to sync with other files.

v2/dtos/common/base.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+// Request defines the base content for request DTOs (data transfer objects).
+// This object and its properties correspond to the BaseRequest object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/BaseRequest
+type BaseRequest struct {
+	RequestID string `json:"requestId" validate:"required,uuid"`
```
cloudxxx8, line 12: @jpwhitemn @iain-anderson I thought the compromised way for now is not set this field as required.
what do you think?

lindseysimple:
Remove the required validation check of the `RequestID` in `BaseRequest` as discussed in the July 2nd core working group meeting.

Issue #254, opened 2020-07-17
DTOs need to explicitly define API version


weichou1229:
https://github.com/edgexfoundry/edgex-go/pull/2610#issuecomment-659076387
Since APP service may receive v1 or v2 event DTO from CoreData, we should specify the API version in the DTO.

PR #255, opened 2020-07-17
feat: Add API version for v2 DTO

weichou1229:
Since APP service may receive v1 or v2 event DTO from CoreData, we should specify the API version in the DTO.
Fix: #254

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #254 

## What is the new behavior?
* Add API version for v2 DTO
* Add func to convert event and reading from model to DTO

- [ ] Yes
- [x] No

## Other information


v2/dtos/reading.go
```diff
@@ -65,3 +67,37 @@ func ToReadingModel(r BaseReading, device string) models.Reading {
 	}
 	return readingModel
 }
+
+func FromReadingModelToDTO(reading models.Reading) BaseReading {
+	var baseReading BaseReading
+	switch reading.(type) {
```
cloudxxx8, line 73: use 
`switch r := reading.(type)`
Then, you don't need to cast in each case
https://tour.golang.org/methods/16

weichou1229: Modified.

v2/dtos/reading.go
```diff
@@ -5,7 +5,9 @@
 
 package dtos
 
-import "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+)
```
lenny-intel, line 10: In Core WG we decided that all DTOs need to be versioned. So please add version to Reading. Make sense that Request DTOs don't need to be versioned.

weichou1229: Version added.

Issue #259, opened 2020-07-31
Update common Response DTOs to have factory methods & remove use of BaseResponse


lenny-intel:
The new swagger definition has removed the BaseResponse from these DTOs.

Also for ease of use each of these Response DTOs  needs a factory method that populates the Response DTO so all services has consistent content in the response, i.e. Timestamp format, apiVersion, etc.

PR #260, opened 2020-07-31
feat: Update common Response DTOs (Add factory method & remove BaseResponse)

lenny-intel:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

BaseResponse in the common DTOs (Ping, config, metrics, version) is no longer needed 
No common factory methods for created the common DTOs .

Issue Number: #259 

## What is the new behavior?

BaseResponse removed from common DTOs.
No common factory methods for created the common DTOs

- [ ] Yes
- [x] No

## Are there any new imports or modules? If so, what are they used for and why?
No
## Other information

### DIFF
--- a/v2/dtos/common/metrics.go
+++ b/v2/dtos/common/metrics.go
@@ -1,15 +1,12 @@
 //
 // Copyright (C) 2020 IOTech Ltd
+// Copyright (C) 2020 Intel Corporation
 //
 // SPDX-License-Identifier: Apache-2.0
 
 package common
 
-// MetricsResponse defines the providing memory and cpu utilization stats of the service.
-// This object and its properties correspond to the MetricsResponse object in the APIv2 specification:
-// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/MetricsResponse
-type MetricsResponse struct {
-	BaseResponse   `json:",inline"`
+type Metrics struct {
 	MemAlloc       uint64 `json:"memAlloc"`
 	MemFrees       uint64 `json:"memFrees"`
 	MemLiveObjects uint64 `json:"memLiveObjects"`
@@ -17,4 +14,20 @@ type MetricsResponse struct {
 	MemSys         uint64 `json:"memSys"`
 	MemTotalAlloc  uint64 `json:"memTotalAlloc"`
 	CpuBusyAvg     uint8  `json:"cpuBusyAvg"`
+}
+
+// MetricsResponse defines the providing memory and cpu utilization stats of the service.
+// This object and its properties correspond to the MetricsResponse object in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/MetricsResponse
+type MetricsResponse struct {
+	Versionable `json:",inline"`
+	Metrics     `json:",inline"`
+}
+
+// NewMetricsResponse creates new MetricsResponse with all fields set appropriately
+func NewMetricsResponse(metrics Metrics) MetricsResponse {
+	return MetricsResponse{
+		Versionable: NewVersionable(),
+		Metrics:     metrics,
+	}
 }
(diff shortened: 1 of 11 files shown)


v2/dtos/common/ping.go
```diff
@@ -1,14 +1,31 @@
...
+)
+
 // PingResponse defines the content of response content for POST Ping DTO
 // This object and its properties correspond to the Ping object in the APIv2 specification:
 // https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/PingResponse
 type PingResponse struct {
-	BaseResponse `json:",inline"`
-	Timestamp    string `json:"timestamp"`
+	Versionable `json:",inline"`
+	Timestamp   string `json:"timestamp"`
```
cloudxxx8, line 20: we should change timestamp to int64 without format, and it would be easier to be accepted by other systems

v2/dtos/common/metrics.go
```diff
@@ -18,3 +23,26 @@ type MetricsResponse struct {
 	MemTotalAlloc  uint64 `json:"memTotalAlloc"`
 	CpuBusyAvg     uint8  `json:"cpuBusyAvg"`
 }
+
+// NewMetricsResponse creates new MetricsResponse with all fields set appropriately
+func NewMetricsResponse(
```
cloudxxx8, line 28: We had better create another Metrics struct for this, because the client code might look like.
m := NewMetricsResponse(1000, 1234, 5647, 2353, 36554, 2654, 56)

It is easy to fill out the wrong number and need to change the function signature if we add more metrics.

```
type Metrics struct {
	MemAlloc       uint64 `json:"memAlloc"`
	MemFrees       uint64 `json:"memFrees"`
	MemLiveObjects uint64 `json:"memLiveObjects"`
	MemMallocs     uint64 `json:"memMallocs"`
	MemSys         uint64 `json:"memSys"`
	MemTotalAlloc  uint64 `json:"memTotalAlloc"`
	CpuBusyAvg     uint8  `json:"cpuBusyAvg"`
}

type MetricsResponse struct {
	Versionable    `json:",inline"`
	Metrics    `json:",inline"`
}

func NewMetricsResponse(m Metrics) MetricsResponse {
...
```

Then, users could call the factory method like this

`m := NewMetricsResponse(Metrics{MemAlloc: 1024, MemFrees: 1000, ....})`

lenny-intel: Yea, I was on the fence about creating the struct, so your feedback has pushed me over it.. ;-) Will create the struct.

v2/dtos/common/ping.go
lenny-intel: This way is human readable and the string is in standard format that can be parsed back to an int64. This is how the swagger has it defined. IMHO, not worth changing since this API is only used by Consul which only looks are the status code... ;-)

v2/dtos/common/metrics.go
lenny-intel: Changed. Much cleaner! Thanks for feedback!

v2/dtos/common/version.go
```diff
@@ -1,21 +1,48 @@
...
+
+// NewVersionSdkResponse creates new VersionSdkResponse with all fields set appropriately
+func NewVersionSdkResponse(appVersion string, sdkVersion string) VersionSdkResponse {
+	return VersionSdkResponse{
+		VersionResponse: VersionResponse{
+			Versionable: Versionable{
+				ApiVersion: v2.ApiVersion,
+			},
+			Version: appVersion},
+
```
eno-intel, line 45: could you call  NewVersionResponse() here, such that the block becomes:
```
VersionResponse: NewVersionResponse(appVersion),
```

lenny-intel: Yes! Excellent idea. I even took it further to Versionable and BaseResponse.

Issue #263, opened 2020-08-07
Add robust validation on the V2 API AddEventRequest & Reading DTOs


lindseysimple:
Per discussion in https://github.com/edgexfoundry/edgex-go/pull/2610#discussion_r466553964, update V2 API core data `AddEventRequest` & `Reading` DTOs by adding the validation annotations to sync with the swagger file definitions.

The latest changes for the swagger file is addressed in https://github.com/edgexfoundry/edgex-go/pull/2653.

PR #264, opened 2020-08-07
Add validation annotations on the V2 core data DTO

lindseysimple:
Add robust validation on the V2 API AddEventRequest & Reading DTOs.

Fix https://github.com/edgexfoundry/go-mod-core-contracts/issues/263

Signed-off-by: Lindsey Cheng <beckysocute@gmail.com>

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #263 

## What is the new behavior?
Validation annotations are added to the required fields of AddEventRequest & Reading DTOs.

- [ ] Yes
- [x] No

## Other information


v2/dtos/event.go
```diff
@@ -18,11 +18,11 @@ type Event struct {
 	common.Versionable `json:",inline"`
 	ID                 string        `json:"id"`
 	Pushed             int64         `json:"pushed,omitempty"`
-	DeviceName         string        `json:"deviceName"`
+	DeviceName         string        `json:"deviceName" validate:"required"`
 	Created            int64         `json:"created"`
 	Modified           int64         `json:"modified,omitempty"`
 	Origin             int64         `json:"origin"`
-	Readings           []BaseReading `json:"readings"`
+	Readings           []BaseReading `json:"readings" validate:"required"`
```
lenny-intel, line 25: I think this needs to be: `validate:"required,gt=0,dive"`

v2/dtos/requests/event.go
```diff
@@ -19,9 +19,7 @@ import (
 // https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-data/2.x#/AddEventRequest
 type AddEventRequest struct {
 	common.BaseRequest `json:",inline"`
-	DeviceName         string             `json:"deviceName" validate:"required"`
-	Origin             int64              `json:"origin" validate:"required"`
-	Readings           []dtos.BaseReading `json:"readings"`
+	Event              dtos.Event `json:"event"`
```
lenny-intel, line 22: Add ` validate:"required,dive"`

v2/dtos/event.go
```diff
@@ -18,11 +18,11 @@ type Event struct {
 	common.Versionable `json:",inline"`
 	ID                 string        `json:"id"`
 	Pushed             int64         `json:"pushed,omitempty"`
-	DeviceName         string        `json:"deviceName"`
+	DeviceName         string        `json:"deviceName" validate:"required"`
 	Created            int64         `json:"created"`
 	Modified           int64         `json:"modified,omitempty"`
 	Origin             int64         `json:"origin"`
```
lenny-intel, line 24: I think  `Origin` be required?

v2/dtos/reading.go
```diff
@@ -18,12 +18,11 @@ import (
 type BaseReading struct {
 	common.Versionable `json:",inline"`
 	Id                 string   `json:"id"`
-	Pushed             int64    `json:"pushed,omitempty"`
 	Created            int64    `json:"created"`
 	Origin             int64    `json:"origin"`
```
lenny-intel, line 22: I think  `Origin` should be required?

v2/dtos/requests/event_test.go
```diff
@@ -21,20 +22,29 @@ var testAddEvent = AddEventRequest{
...
 	noReID := testAddEvent
 	noReID.RequestID = ""
 	noDeviceName := testAddEvent
-	noDeviceName.DeviceName = ""
+	noDeviceName.Event.DeviceName = ""
 	noDeviceOrigin := testAddEvent
-	noDeviceOrigin.DeviceName = ""
-	noDeviceOrigin.Origin = 0
+	noDeviceOrigin.Event.DeviceName = ""
+	noDeviceOrigin.Event.Origin = 0
```
lenny-intel, line 47: Add tests cases for missing Origin, missing Readings and invalid Reading (missing DeviceName, missing Name & missing  Origin)

v2/dtos/event.go
lenny-intel: FYI, @tonyespy and I will bring this up in the Device Services WG meeting on Monday. I have verified that V1 doesn't require `Origin` is set, but IMHO that is a miss in V1, since if the device doesn't set it the Device Service does.

lenny-intel: In the interest in wrapping up `feat: Add POST event functions based for v2 API` edgexfoundry/edgex-go#2610. Leave `Origin` as is and we can deal with in a follow-up issue/pr.

lenny-intel:
@cloudxxx8 , if you feel my comments have been addressed please feel free to approve and merge so @lindseysimple can update edgexfoundry/edgex-go#2610 for final approval and merge on my Monday morning. THX!

v2/dtos/reading.go
```diff
@@ -18,12 +18,11 @@ import (
...
 	Modified           int64    `json:"modified,omitempty"`
-	DeviceName         string   `json:"deviceName"`
-	Name               string   `json:"name"`
+	DeviceName         string   `json:"deviceName" validate:"required"`
+	Name               string   `json:"name" validate:"required"`
 	Labels             []string `json:"labels,omitempty"`
 	BinaryReading      `json:",inline"`
```
lenny-intel, line 27: `FloatEncoding` is required if `ValueType` = `ValueTypeFloat32` or  `ValueTypeFloat64`

lenny-intel: Also should `ValueType` be validated for accepted values? That is a long list and not done in V1's validation. Just a question that we should consider.

v2/dtos/event.go
tonyespy: Why does `Modified` have "omitempty" but `Created` does not?

v2/dtos/reading.go
tonyespy: Why does `Modified` have "omitempty" but `Created` does not?

v2/dtos/event.go
cloudxxx8: `Modified `have "omitempty" because it is empty before updating
`Created `will always contain value once it is created
However, I thought we can remove `Modified` from `Event `and `Reading`.  We had it because it was in `DescribableObject`.
`Event `and `Reading `cannot be updated, and `Pushed` is enough for `Event`

v2/dtos/event.go
lindseysimple: Updated.

v2/dtos/requests/event.go
lindseysimple: Thanks. Updated to `validate:"required"`. No need for the `dive` tag here as the inner `Event` is not a slice, array or map.

v2/dtos/requests/event_test.go
lindseysimple: Added as suggested.

v2/dtos/reading.go
lindseysimple: > `FloatEncoding` is required if `ValueType` = `ValueTypeFloat32` or `ValueTypeFloat64`

Added the check in `func (b BaseReading) Validate()`.

v2/dtos/event.go
lindseysimple: Remove `Modified` field from `Event` and `Reading` DTOs/models as @cloudxxx8 suggested.

Updated `Origin` as required and also applied in PR https://github.com/edgexfoundry/edgex-go/pull/2653.

v2/dtos/reading.go
```diff
@@ -47,6 +52,33 @@ type SimpleReading struct {
 	FloatEncoding string `json:"floatEncoding,omitempty"`
 }
 
+// Validate satisfies the Validator interface
+func (b BaseReading) Validate() error {
+	// check if the inner BinaryReading struct is empty
+	if !reflect.DeepEqual(b.BinaryReading, BinaryReading{}) {
```
lenny-intel, line 58: Can't we determine this based on ValueType, which should be common to both Binary & Simple, i.e should be in BaseReading. This will be an issue with consumers of the DTO's, i.e.App services, if we can't do this is a simple check of the ValueType.

v2/dtos/requests/event.go
```diff
@@ -34,25 +32,30 @@ type UpdateEventPushedByChecksumRequest struct {
...
 // Validate satisfies the Validator interface
 func (a AddEventRequest) Validate() error {
-	err := v2.Validate(a)
-	return err
+	if err := v2.Validate(a); err != nil {
+		return err
+	}
+
+	// validate the nested BinaryReading and SimpleReading struct
+	for _, r := range a.Event.Readings {
```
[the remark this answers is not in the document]
lindseysimple: Yes, originally it would actually check the nested structs with the following annotation added:
https://github.com/edgexfoundry/go-mod-core-contracts/pull/264/files/a67534a7810fd718be6ef5ae376fb7cd36490ba5#diff-7a43efa7e1bef0ec66b9fa81dddddf11R24

However, as we only have either `SimpleReading` or `BinaryReading` in `BaseReading` DTO, the above check would cause the validation failed so the we do the following to skip the validation of `SimpleReading` and `BinaryReading`:
https://github.com/edgexfoundry/go-mod-core-contracts/pull/264/files/a67534a7810fd718be6ef5ae376fb7cd36490ba5#diff-badcd41a51a415996619cb57beb0c5c5R33-R34

Therefore additional handling is added to handle the nested validation of `SimpleReading` and `BinaryReading`:
https://github.com/edgexfoundry/go-mod-core-contracts/pull/264/files/a67534a7810fd718be6ef5ae376fb7cd36490ba5#diff-fd093d7d5d89b1cdab65572c38e1b4baR39-R42

Would you please suggest if any other workaround you think of?? Thanks.

v2/dtos/reading.go
lindseysimple: Updated `Origin` as required in `BaseReading` DTO.

v2/dtos/reading.go
cloudxxx8: No, @lenny-intel , BinaryReading only contains mediaType, not ValueType.
If we do so, we need to move ValueType to BaseReading.  Is it ok?

lenny-intel: Yes, IMO, `ValueType` has to be in `BaseReading`. It is one of the key fields that App Services can filter one.

v2/dtos/requests/event.go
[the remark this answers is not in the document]
lenny-intel: Ok, got it, if it is a `SimpleReading` the validation fails for the `BinaryReading` when using annotation.
Please comment `why doing this` in the code. THX!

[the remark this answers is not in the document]
lindseysimple: Comments added, thanks.

v2/dtos/reading.go
lindseysimple: Updated to define `ValueType` in `BaseReading` instead of `SimpleReading.`

v2/dtos/reading.go
```diff
@@ -43,23 +49,50 @@ type BinaryReading struct {
...
 type SimpleReading struct {
 	Value         string `json:"value" validate:"required"`
-	ValueType     string `json:"valueType" validate:"required"`
 	FloatEncoding string `json:"floatEncoding,omitempty"`
 }
 
+// Validate satisfies the Validator interface
+func (b BaseReading) Validate() error {
+	// check if the inner BinaryReading struct is empty
+	if !reflect.DeepEqual(b.BinaryReading, BinaryReading{}) {
```
lenny-intel, line 58: Now, you should be able to check `ValueType == ValueTypeBinary`  rather than do this Deep Equal reflection check. If not Binary, assume `SimpleReading`. i.e. no need to final `else` on line 75

v2/dtos/requests/event_test.go
```diff
@@ -21,20 +22,116 @@ var testAddEvent = AddEventRequest{
...
+			Value: TestReadingFloatValue,
+		},
+	}}
+
+	invalidBinaryReadingNoValue := testAddEvent
+	invalidBinaryReadingNoValue.Event.Readings = []dtos.BaseReading{{
+		DeviceName: TestDeviceName,
+		Name:       TestDeviceResourceName,
+		Origin:     TestOriginTime,
+		BinaryReading: dtos.BinaryReading{
```
lenny-intel, line 119: Need `ValueType: ValueTypeBinary` here and in all the Binary test cases. These should have failed validation on `ValueType` required.

v2/dtos/reading.go
lindseysimple: Updated to check inner struct by `ValueType` instead of reflect, thanks.

v2/dtos/requests/event_test.go
lindseysimple: Updated the Binary test cases for adding `ValueType`. Thanks for catching that.

PR #288, opened 2020-09-13
Feat (core-data): Event ID has to be UUID and pre-populated

### LINKED ISSUE #287: Pre-populate the event ID with a UUID
copied the description from related to https://github.com/edgexfoundry/edgex-go/issues/2684
we need to start from the modification in contracts

# 🚀 Feature Request
Make ID field of event required and provided before passing to core data for persistence.

### Relevant Package
core data and core contracts

### Description
This feature began as an ADR (see https://github.com/edgexfoundry/edgex-docs/pull/227).  Per the discussion, the idea is to 
generate a UUID and assign it to the id field before storing the event to the database. If the id field was empty at the time of persistence, then have core data generate a UUID (or just have the ID field as required making clients provide the UUID in the ID).

Per @lenny-intel - this also solves why we used checksum in the first place for CBOR encoded events. Now the ID will be set in the CBOR encoded Event and accessible in App Service once it is un-marshaled. Then we only need to have the MarkAsPushedById API.

cloudxxx8:
## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Event ID can be empty before sending to core-data and assigned in core-data

Issue Number: Fix #287 

## What is the new behavior?
Validate Event ID is an UUID and pre-populated

- [ ] Yes
- [x] No

## Are there any new imports or modules? If so, what are they used for and why?
No

## Are there any specific instructions or things that should be known prior to reviewing?
No

### DIFF
--- a/models/event_test.go
+++ b/models/event_test.go
@@ -25,6 +25,7 @@ import (
 )
 
 var TestEvent = Event{
+	ID:       "1e12bd0a-89ca-4747-ad76-a43157d6521a",
 	Pushed:   123,
 	Created:  123,
 	Device:   TestDeviceName,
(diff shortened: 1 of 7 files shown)


lenny-intel (commented):
LGTM, but do we also need to do the same for Reading Id?

cloudxxx8:
recheck

cloudxxx8:
> LGTM, but do we also need to do the same for Reading Id?

no, I don't think so.  the purpose of this design is to replace checksum for "pushed" API.
I can't think out any advantage to do the same in Reading.

lenny-intel:
> I can't think out any advantage to do the same in Reading.

If core data doesn't persist or is bypassed then the reading ID is never set.

cloudxxx8:
reading ID is for database identification.  if readings are not persisted, they will always with Event, and ID is not a matter.

lenny-intel:
> ID is not a matter.

Ok, ya sold me. ;-)

lenny-intel (approved):

PR #307, opened 2020-09-26
fix(v2dto):Remove validate call in add request DTO 

### LINKED ISSUE #306: V2 DTO AddRequests UnmarshalJSON() should not execute Validate()
According to the V2 API design, Add and Update related requests are arrays of requests, and it should still process all the requests if some of the request cannot pass the validation.  
In the current implementation, unmarshal the request array would validate all of the requests and return the error even if some of the requests are valid.
The `Validate()` func should be invoked by the Controller in edgex-go individually.

cloudxxx8:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: Fix #306

## What is the new behavior?
In V2 API design, all the add request would be in array form when
calling add APIs.  Thus, the requests should be validated
individually, not in the UnmarshalJSON func.

- [ ] Yes
- [x] No

### DIFF
--- a/v2/dtos/deviceprofile.go
+++ b/v2/dtos/deviceprofile.go
@@ -27,7 +27,7 @@ type DeviceProfile struct {
 	CoreCommands       []Command         `json:"coreCommands,omitempty" yaml:"coreCommands,omitempty" validate:"dive"`
 }
 
-// ToCommandModels transforms the Command DTOs to the Command models
+// ToDeviceProfileModels transforms the DeviceProfile DTOs to the DeviceProfile models
 func ToDeviceProfileModels(deviceProfileDTO DeviceProfile) models.DeviceProfile {
 	return models.DeviceProfile{
 		Name:            deviceProfileDTO.Name,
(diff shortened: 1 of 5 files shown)


lenny-intel (commented):
Another option would be for these functions to return []error and have them iterate over the requests to validate each one and build the slice. This would keep validation in the DTO, rather than push it to the Controllers.

lenny-intel (commented):
The controllers would have to change to call theses functions directly.

cloudxxx8:
> Another option would be for these functions to return []error and have them iterate over the requests to validate each one and build the slice. This would keep validation in the DTO, rather than push it to the Controllers.

it won't work because the DTO is a single request, not an array of requests.
unless we create another DTO to contain an array of add requests

lenny-intel:
> it won't work because the DTO is a single request, not an array of requests.
@cloudxxx8 Right, I meant to iterate over the array of profiles in the request. We still need to validate the BaseRequest. The reason I would like to keep the validation here is so that it is done for any protocol. If moved to the HTTP Controller, we then have to repeat the validation in future protocol controllers.

Other thought I have is yes, we handle multiple profiles but the validation is for the whole request. I.e. we can handle errors processing individual profiles and report back status for each, but the request as a whole needs to be valid.

lenny-intel:
@cloudxxx8 , after further review I don't agree with removing the validation here. Good topic for tomorrows Core WG.

PR #336, opened 2020-10-14
feat(v2 api): Add new error kind for indicating requested range not satisfiable

### LINKED ISSUE #334: [V2 API]  Add new error kind for indicating requested range not satisfiable
As V2 API supports the pagination through offset and limit query string when GET multiple objects, there is a need to add new error kind when the requested range is not satisfiable.  For example, when there are only 5 objects in the DB but users requests to retrieve objects starting from 10 (offset=9), V2 API shall return an error kind that such requested range is not available.  This new error kind shall also map to HTTP status code 416 as indicated in the [RFC 7233](https://tools.ietf.org/html/rfc7233#section-4.4).

judehung:
As V2 API supports the pagination through offset and limit query string
when GET multiple objects, there is a need to add new error kind when
the requested range is not satisfiable. For example, when there are only
5 objects in the DB but users requests to retrieve objects starting from
10 (offset=9), V2 API shall return an error kind that such requested
range is not available. This new error kind shall also map to HTTP status
code 416 as indicated in the RFC 7233.

Fix #334 

Signed-off-by: Jude Hung <jude@iotechsys.com>

## PR Checklist

## Issue Number:
#334 

## What is the new behavior?
A new error kind `KindRangeNotSatisfiable` is added for indicating requested range not satisfiable.

- [ ] Yes
- [x] No

## New Imports
Are there any new imports or modules? If so, what are they used for and why?
No

## Specific Instructions
https://tools.ietf.org/html/rfc7233#section-4.4

## Other information

### DIFF
--- a/errors/types.go
+++ b/errors/types.go
@@ -17,19 +17,20 @@ type ErrKind string
 
 const (
 	// Constant Kind identifiers which can be used to label and group errors.
-	KindUnknown            ErrKind = "Unknown"
-	KindDatabaseError      ErrKind = "Database"
-	KindCommunicationError ErrKind = "Communication"
-	KindEntityDoesNotExist ErrKind = "NotFound"
-	KindContractInvalid    ErrKind = "ContractInvalid"
-	KindServerError        ErrKind = "UnexpectedServerError"
-	KindLimitExceeded      ErrKind = "LimitExceeded"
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

Issue #342, opened 2020-10-15
[V2] Remove xxResponseNoMessage helper functions


cloudxxx8:
In 15-OCT-2020 Core WG meeting, we decided to remove all the xxResponseNoMessage helper functions to reduce the future development and unit test effort.

PR #345, opened 2020-10-18
refactor: remove all ResponseNoMessage funcs in V2

cloudxxx8:
## PR Checklist

## Issue Number:
Fix https://github.com/edgexfoundry/go-mod-core-contracts/issues/342

## What is the new behavior?
In 15-OCT-2020 Core WG meeting, we decided to remove all the xxResponseNoMessage helper functions to reduce the future development and unit test effort.
Users could call NewXXResponse with empty string in the message
peremeter instead.

- [ ] Yes
- [x] No

### DIFF
--- a/v2/dtos/responses/event_test.go
+++ b/v2/dtos/responses/event_test.go
@@ -28,19 +28,6 @@ func TestNewEventCountResponse(t *testing.T) {
 	assert.Equal(t, expectedDeviceName, actual.DeviceName)
 }
 
-func TestNewEventCountResponseNoMessage(t *testing.T) {
-	expectedRequestId := "123456"
-	expectedStatusCode := 200
-	expectedCount := uint32(1000)
-	expectedDeviceName := "device1"
-	actual := NewEventCountResponseNoMessage(expectedRequestId, expectedStatusCode, expectedCount, expectedDeviceName)
-
-	assert.Equal(t, expectedRequestId, actual.RequestId)
-	assert.Equal(t, expectedStatusCode, actual.StatusCode)
-	assert.Equal(t, expectedCount, actual.Count)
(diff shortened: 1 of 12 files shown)


v2/dtos/responses/event_test.go
```diff
@@ -80,27 +56,3 @@ func TestNewMultiEventsResponse(t *testing.T) {
...
-	expectedStatusCode := 200
-	expectedMessage := "unit test message"
-	expectedChecksum := "04698a6f20feecb8bbf7cd01e59d31ba1ce17b24ba14b71a8fb370065d951f57"
-	actual := NewUpdateEventPushedByChecksumResponse(expectedRequestId, expectedMessage, expectedStatusCode, expectedChecksum)
-
-	assert.Equal(t, expectedRequestId, actual.RequestId)
-	assert.Equal(t, expectedStatusCode, actual.StatusCode)
-	assert.Equal(t, expectedMessage, actual.Message)
-	assert.Equal(t, expectedChecksum, actual.Checksum)
-}
```
lenny-intel, line 95: This isn't a `NoMessageResponse` test. Why removing?

I do think it will go way for another reason, which should be in a separate PR. That reason is with the new ID required and set from start we will no longer need checksum. So the EventPushedByChecksum API should also go away.

lenny-intel (changes requested):

v2/dtos/responses/event_test.go
cloudxxx8: it's a mistake.  I didn't have intention to remove this...
added it back.  thanks for the catching

cloudxxx8:
rebased

lenny-intel (approved):
LGTM

Issue #309, opened 2020-09-29
V2 API models contain incorrect or empty description


cherrycl:
There are some models with no parameter description, like [deviceresource.go](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/v2/models/deviceresource.go).
Some models with incorrect descriptions, like `LastReported` in [deviceservice.go](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/v2/models/deviceservice.go#L16).

cloudxxx8:
@lenny-intel @jpwhitemn @iain-anderson 
We had better put these field description into a single place, such as Swagger file.
In current go source code, some of them contain full description, but some of them don't

Full example
https://github.com/edgexfoundry/go-mod-core-contracts/blob/a48a3cfc95ac7623bae8df3d1eb5020ec58becf8/v2/models/device.go#L8-L25

Incorrect example
https://github.com/edgexfoundry/go-mod-core-contracts/blob/a48a3cfc95ac7623bae8df3d1eb5020ec58becf8/v2/models/deviceservice.go#L16

Not comment example
https://github.com/edgexfoundry/go-mod-core-contracts/blob/a48a3cfc95ac7623bae8df3d1eb5020ec58becf8/v2/models/deviceresource.go#L9-L15

I suggest to remove all them from source code, and only leave the comment if the fields are not the same as DTO.

lenny-intel:
Yes, I like putting the descriptions in only the Swagger and and comment to the structs that says to see swagger for details.

PR #346, opened 2020-10-18
docs: Remove docs in V2 model and link to swagger

cloudxxx8:
## PR Checklist

There are some inconsistent docs in V2 model and dto

## Issue Number:
Fix https://github.com/edgexfoundry/go-mod-core-contracts/issues/309

## What is the new behavior?
We point to SwaggerHub which is the API documentation.  We will use the single of
source for the documentation.

- [ ] Yes
- [x] No


lenny-intel (changes requested):
I am good with the link to swagger for the DTOs, but not for the models. The swagger documents the DTOs, not the model's.

cloudxxx8:
@lenny-intel that's the original discussion in the #309, and I mean model in that PR.
Should we copy all the same description to the model package?  So far, they are the same.
My suggestion is to link to the swagger file and write own the comment for the difference only.

lenny-intel:
> My suggestion is to link to the swagger file and write own the comment for the difference only.

I am ok with that approach if we add the comment for each model like:
`Model fields are same as the DTOs documented by this swagger. Exceptions, if any, are noted below.`

cloudxxx8:
> > My suggestion is to link to the swagger file and write own the comment for the difference only.
> 
> I am ok with that approach if we add the comment for each model like:
> `Model fields are same as the DTOs documented by this swagger. Exceptions, if any, are noted below.`

added

lenny-intel (approved):
LGTM

Issue #343, opened 2020-10-15
[V2] Make sure the name field only contains the unreserved characters in RFC3986


cloudxxx8:
In 15-OCT-2020 Core WG meeting, we decided to use only unreserved characters in the name field according to RFC3986.

https://en.wikipedia.org/wiki/Percent-encoding

weichou1229:
This tag might help:
https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-URL_Encoded

> This validates that a string value is percent-encoded (URL encoded) according to https://tools.ietf.org/html/rfc3986#section-2.1

judehung:
> This tag might help:
> https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-URL_Encoded
> 
> > This validates that a string value is percent-encoded (URL encoded) according to https://tools.ietf.org/html/rfc3986#section-2.1

@weichou1229 thanks for the information.  Unfortunately, ulr_encoded tag is used to validate if a string is url-encoded or not, which is a little bit different to what we would like to validate.  I'll implement a custom tag to validate that name field only allows  RFC3986 section 2.3 defined unreserved characters, which should be `ALPHA / DIGIT / "-" / "." / "_" / "~"`.

PR #363, opened 2020-10-26
fix(metadata): Add validation tag for UpdateDTO

weichou1229:
## PR Checklist

- Update device service with empty name will return status code 404
- It does not allow to update the baseAddress with empty string.

## What is the new behavior?
UpdateDeviceDTO and UpdateDeviceServiceDTO should be able to verify the id whether is valid UUID format and the required name should not be empty.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions
I tried several tag combinations, like,
- required_without and uuid tag
- required_without and gt=0 tag
- required_without and omitempty and uuid tag
- required_without and omitempty and gt=0 tag

But it didn't pass expected test cases:
- When Id exists, Id should be valid UUID
- When Name exists, Name should not be empty

So I try to write tags  **edgex-update-dto-id**, **edgex-update-dto-id** to verify Id and Name.

Please help review whether the implementation is properly.

### DIFF
--- a/v2/dtos/device.go
+++ b/v2/dtos/device.go
@@ -14,10 +14,10 @@ import (
 // https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-metadata/2.x#/Device
 type Device struct {
 	common.Versionable `json:",inline"`
-	Id                 string                        `json:"id,omitempty"`
+	Id                 string                        `json:"id,omitempty" validate:"omitempty,uuid"`
 	Created            int64                         `json:"created,omitempty"`
 	Modified           int64                         `json:"modified,omitempty"`
-	Name               string                        `json:"name" validate:"required"`
+	Name               string                        `json:"name" validate:"required,edgex-dto-none-empty-string"`
 	Description        string                        `json:"description,omitempty"`
 	AdminState         string                        `json:"adminState" validate:"oneof='LOCKED' 'UNLOCKED'"`
 	OperatingState     string                        `json:"operatingState" validate:"oneof='ENABLED' 'DISABLED'"`
@@ -25,8 +25,8 @@ type Device struct {
 	LastReported       int64                         `json:"lastReported,omitempty"`
 	Labels             []string                      `json:"labels,omitempty"`
 	Location           interface{}                   `json:"location,omitempty"`
-	ServiceName        string                        `json:"serviceName" validate:"required"`
-	ProfileName        string                        `json:"profileName" validate:"required"`
+	ServiceName        string                        `json:"serviceName" validate:"required,edgex-dto-none-empty-string"`
+	ProfileName        string                        `json:"profileName" validate:"required,edgex-dto-none-empty-string"`
 	AutoEvents         []AutoEvent                   `json:"autoEvents,omitempty" validate:"dive"`
 	Protocols          map[string]ProtocolProperties `json:"protocols,omitempty" validate:"required,gt=0"`
 }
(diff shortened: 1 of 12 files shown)


v2/validator.go
```diff
@@ -73,3 +79,29 @@ func ValidateAutoEventFrequency(fl validator.FieldLevel) bool {
...
+	if val.Kind() != reflect.String {
+		return true
+	}
+	// The string value should not be empty
+	if len(strings.TrimSpace(val.String())) > 0 {
+		return true
+	} else {
+		return false
+	}
+}
```
lenny-intel, line 107: Since this are `*string` types, should be checking `val` for `nil` first?

And ` val.Kind() != reflect.String` will always be true since the type is `*string`.

v2/dtos/deviceservice.go
```diff
@@ -30,9 +30,9 @@ type DeviceService struct {
 // UpdateDeviceService and its properties are defined in the APIv2 specification:
 // https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-metadata/2.x#/UpdateDeviceService
 type UpdateDeviceService struct {
-	Id             *string  `json:"id" validate:"required_without=Name"`
-	Name           *string  `json:"name" validate:"required_without=Id"`
-	BaseAddress    *string  `json:"baseAddress"`
+	Id             *string  `json:"id" validate:"required_without=Name,edgex-update-dto-id"`
+	Name           *string  `json:"name" validate:"required_without=Id,edgex-update-dto-name"`
+	BaseAddress    *string  `json:"baseAddress" validate:"omitempty,gt=0"`
```
lenny-intel, line 35: Why is this needed? what does `gt=0` mean for a `*string` type? Is this way to validate that it isn't an empty string? What about blanks. BaseAddress is a fully qualified URL. Should validate it is a fully qualified URL, it not nil.

v2/dtos/device.go
```diff
@@ -34,15 +34,15 @@ type Device struct {
...
+	Name           *string                       `json:"name" validate:"required_without=Id,edgex-update-dto-name"`
+	Description    *string                       `json:"description" validate:"omitempty,gt=0"`
 	AdminState     *string                       `json:"adminState" validate:"omitempty,oneof='LOCKED' 'UNLOCKED'"`
 	OperatingState *string                       `json:"operatingState" validate:"omitempty,oneof='ENABLED' 'DISABLED'"`
 	LastConnected  *int64                        `json:"lastConnected"`
 	LastReported   *int64                        `json:"lastReported"`
-	ServiceName    *string                       `json:"serviceName"`
-	ProfileName    *string                       `json:"profileName"`
+	ServiceName    *string                       `json:"serviceName" validate:"omitempty,gt=0"`
+	ProfileName    *string                       `json:"profileName" validate:"omitempty,gt=0"`
```
lenny-intel, line 45: Why not use `edgex-update-dto-name` instead of `gt=0` so string of blanks is caught.

v2/validator.go
```diff
@@ -73,3 +79,29 @@ func ValidateAutoEventFrequency(fl validator.FieldLevel) bool {
...
+	// Since the Id should be string type, so skip the validation when it isn't a kind of string
+	if val.Kind() != reflect.String {
+		return true
+	}
+	_, err := uuid.Parse(val.String())
+	return err == nil
+}
+
+// ValidateUpdateDtoName used to check the UpdateDTO name pointer value
+func ValidateUpdateDtoName(fl validator.FieldLevel) bool {
```
lenny-intel, line 95: I like this and see that it can be expanded to validate the `name` only have valid characters.

lenny-intel (changes requested):

v2/validator.go
cloudxxx8: ok, we will add the validation for name when resolving https://github.com/edgexfoundry/go-mod-core-contracts/issues/343

v2/validator.go
weichou1229: > Since this are *string types, should be checking val for nil first?

Yes, I modified the code to check the nil pointer.
```
if val.Kind() == reflect.Ptr && val.IsNil() 
```

> And val.Kind() != reflect.String will always be true since the type is *string

IF the pointer value is nil, then val.Kind() => ptr
IF the pointer value is string, then val.kind() => string

v2/dtos/deviceservice.go
weichou1229: > Why is this needed? what does gt=0 mean for a *string type? Is this way to validate that it isn't an empty string? What about blanks. 

The gt=0 is used to check string length, but we should check the blank string as you mentioned.

> BaseAddress is a fully qualified URL. Should validate it is a fully qualified URL, it not nil.

Added uri tag for validation. https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-URI_String

v2/dtos/device.go
weichou1229: Modified. You're right, we should check the blank string.

lenny-intel (approved):
LGTM

PR #443, opened 2021-01-20
feat(command): Revise CommandResponse DTO to MultiCoreCommandsResponse

### LINKED ISSUE #441: [V2 command]  Revise CommandResponse DTO to MultiCoreCommandsResponse
https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-command/2.x#/CommandResponse defines array of CommandResponse DTO as the response for GET /device/all and single CommnadResponse DTO as response for GET /device/name/{name}.  After discussion with Cloud, this structure could be simplified by moving device field into CoreCommand DTO, and then both GET /device/all and GET /device/name/{name} could have single CommandResponse with multiple CoreCommand DTOs.  Moreover, with such change, the CommandResponse DTO is suggested to renamed as MultiCoreCommandsResponse.

judehung:
Signed-off-by: Jude Hung <jude@iotechsys.com>

## PR Checklist

https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-command/2.x#/CommandResponse defines array of CommandResponse DTO as the response for GET /device/all and single CommnadResponse DTO as response for GET /device/name/{name}. After discussion with Cloud, this structure could be simplified by moving device field into CoreCommand DTO, and then both GET /device/all and GET /device/name/{name} could have single CommandResponse with multiple CoreCommand DTOs. Moreover, with such change, the CommandResponse DTO is suggested to renamed to MultiCoreCommandsResponse.

## Issue Number:
fix #441 

## What is the new behavior?
1. Create and rename CommandResposne DTO to MultiCoreCommandsResponse
2. Create CoreCommand DTO

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

### DIFF
--- a/v2/dtos/corecommand.go
+++ b/v2/dtos/corecommand.go
@@ -10,8 +10,8 @@ package dtos
 type CoreCommand struct {
 	Name       string `json:"name" validate:"required,edgex-dto-none-empty-string,edgex-dto-rfc3986-unreserved-chars"`
 	DeviceName string `json:"deviceName" validate:"required,edgex-dto-rfc3986-unreserved-chars"`
-	Get        bool   `json:"get,omitempty" validate:"required_without=Put"`
-	Put        bool   `json:"put,omitempty" validate:"required_without=Get"`
+	Get        bool   `json:"get" validate:"required_without=Put"`
+	Put        bool   `json:"put" validate:"required_without=Get"`
 	Path       string `json:"path,omitempty"`
 	Url        string `json:"url,omitempty"`
 }
--- a/v2/dtos/corecommand.go
+++ b/v2/dtos/corecommand.go
@@ -1,0 +1,17 @@
+//
+// Copyright (C) 2021 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package dtos
+
+// CoreCommand and its properties are defined in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-command/2.x#/CoreCommand
+type CoreCommand struct {
+	Name       string `json:"name" validate:"required,edgex-dto-none-empty-string,edgex-dto-rfc3986-unreserved-chars"`
+	DeviceName string `json:"deviceName" validate:"required,edgex-dto-rfc3986-unreserved-chars"`
+	Get        bool   `json:"get,omitempty" validate:"required_without=Put"`
+	Put        bool   `json:"put,omitempty" validate:"required_without=Get"`
+	Path       string `json:"path,omitempty"`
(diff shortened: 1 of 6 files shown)


v2/dtos/corecommand.go
```diff
@@ -0,0 +1,17 @@
...
+
+package dtos
+
+// CoreCommand and its properties are defined in the APIv2 specification:
+// https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-command/2.x#/CoreCommand
+type CoreCommand struct {
+	Name       string `json:"name" validate:"required,edgex-dto-none-empty-string,edgex-dto-rfc3986-unreserved-chars"`
+	DeviceName string `json:"deviceName" validate:"required,edgex-dto-rfc3986-unreserved-chars"`
+	Get        bool   `json:"get,omitempty" validate:"required_without=Put"`
+	Put        bool   `json:"put,omitempty" validate:"required_without=Get"`
```
cloudxxx8, line 14: Get and Put should be both required, and they can not be both false.
This validation rule also applies to Command DTO

cloudxxx8 (changes requested):

v2/dtos/corecommand.go
judehung: Okay, after looking into https://pkg.go.dev/gopkg.in/go-playground/validator.v9, I think we need a custom tag to achieve this validation.

cloudxxx8: We don't have to create a custom tag for this, since it is an easy check and might not be reused.
Please just add the check in its Validate() func like this:
https://github.com/edgexfoundry/go-mod-core-contracts/blob/1915b16eb2eb9b1cb835de2bb0cf2b3eddccddfb/v2/dtos/requests/deviceprofile.go#L28-L34

judehung: Per my experimentation and discussion, we actually don't have to add extra codes to validate false GET and false PUT case, as `required_without` tag could actually satisfy our validation need.

Note:  the [go doc for required_without](https://pkg.go.dev/gopkg.in/go-playground/validator.v9#hdr-Required_Without) is misleading, and should be revised to: 
```
The field under validation must be not the data types default zero value and not empty only when any of the other specified fields are not the data types default zero value and not empty.
```

GET and PUT are defined as bool, which has false as default value, so that following struct could actually throw validation error from go validator when Get=false and Put=false.
```
Get        bool   `json:"get" validate:"required_without=Put"`
Put        bool   `json:"put" validate:"required_without=Get"`
```

judehung:
@cloudxxx8 as this PR includes 3 commits, remember to squash and merge when your review result is postive.

cloudxxx8 (approved):
LGTM

PR #462, opened 2021-01-26
feat(v2): Http client lib set API version to request body

### LINKED ISSUE #450: [V2] Create factory methods for Request DTOs
We have factory methods for Response DTOs, and we also need them for Request DTOs.
Those methods should assign the correct ApiVersion.

weichou1229:
Close #450

## PR Checklist

## What is the new behavior?
Http client set API version to request body for POST, PUT, PATCH operation

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

Use golang reflect to set the API version instead of creating factory methods.

## Other information

### DIFF
--- a/v2/clients/http/utils/request.go
+++ b/v2/clients/http/utils/request.go
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2020 IOTech Ltd
+// Copyright (C) 2020-2021 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -10,8 +10,10 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/url"
+	"reflect"
 
 	"github.com/edgexfoundry/go-mod-core-contracts/v2/errors"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/v2"
 )
 
 // Helper method to make the get request and return the body
@@ -31,13 +33,33 @@ func GetRequest(ctx context.Context, returnValuePointer interface{}, baseUrl str
 	return nil
 }
 
+func setApiVersion(data interface{}) {
+	val := reflect.ValueOf(data).Elem()
+	switch val.Kind() {
+	case reflect.Struct:
+		for i := 0; i < val.NumField(); i++ {
+			if val.Type().Field(i).Name == v2.ApiVersionField {
+				val.Field(i).SetString(v2.ApiVersion)
+			} else {
+				field := val.Field(i).Addr().Interface()
+				setApiVersion(field)
+			}
+		}
+	case reflect.Slice:
+		for i := 0; i < val.Len(); i += 1 {
+			ele := val.Index(i).Addr().Interface()
+			setApiVersion(ele)
+		}
+	}
+}
+
 // Helper method to make the post JSON request and return the body
 func PostRequest(
 	ctx context.Context,
 	returnValuePointer interface{},
 	url string,
 	data interface{}) errors.EdgeX {
-
+	setApiVersion(data)
 	req, err := createRequestWithRawData(ctx, http.MethodPost, url, data)
 	if err != nil {
 		return errors.NewCommonEdgeXWrapper(err)
@@ -59,7 +81,7 @@ func PutRequest(
 	returnValuePointer interface{},
 	url string,
 	data interface{}) errors.EdgeX {
-
+	setApiVersion(data)
 	req, err := createRequestWithRawData(ctx, http.MethodPut, url, data)
 	if err != nil {
 		return errors.NewCommonEdgeXWrapper(err)
@@ -81,7 +103,7 @@ func PatchRequest(
 	returnValuePointer interface{},
 	url string,
 	data interface{}) errors.EdgeX {
-
+	setApiVersion(data)
 	req, err := createRequestWithRawData(ctx, http.MethodPatch, url, data)
 	if err != nil {
 		return errors.NewCommonEdgeXWrapper(err)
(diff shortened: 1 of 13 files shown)


v2/dtos/requests/event.go
```diff
@@ -89,3 +89,14 @@ func AddEventReqToEventModel(addEventReq AddEventRequest) (event models.Event) {
 		Tags:        tags,
 	}
 }
+
+func NewAddEventRequest(requestId string, event dtos.Event) AddEventRequest {
+	event.Versionable = common.NewVersionable()
+	for _, reading := range event.Readings {
+		reading.Versionable = common.NewVersionable()
```
lenny-intel, line 96: In this case `reading` is a copy, so you must use the `index` to modify properties on the reading for change to stick outside the current scope of `reading`

v2/clients/http/utils/request.go
```diff
@@ -31,13 +33,33 @@ func GetRequest(ctx context.Context, returnValuePointer interface{}, baseUrl str
 	return nil
 }
 
+func setApiVersion(data interface{}) {
```
lenny-intel, line 36: Why do we need this when we have added the New factory method. I assume there will be new factory methods for the other Requests for consistency?

lenny-intel (changes requested):

v2/dtos/requests/event.go
weichou1229: Thank you for catching that!   Modified.

v2/clients/http/utils/request.go
weichou1229: Our purpose is to prevent sending POST, PUT, PATCH request without API version.
- **NewAddEventRequest**: Services need to use NewAddEventRequest to create a AddEventRequest and send the request via message bus.
- **setApiVersion**: Otherwise we can send request via client library and the setApiVersion func can help to add the api version.

lenny-intel: I would rather see ApiVersion as a required field rather than doing reflection to make sure it is set. i.e. if this is the need, why not use validation like on other fields for make sure it is set properly.

@cloudxxx8 , thoughts?

cloudxxx8: @lenny-intel I agree with setting ApiVersion as a required field, but we can still make sure everything using the client library would set the ApiVersion automatically.  If the users use V2 client library, they are using V2 API, not other version.

I changed my mind about creating new factory methods for request, because there will be lots of input parameters, much more than response.  It would be hard to use.

lenny-intel: > I changed my mind about creating new factory methods for request, because there will be lots of input parameters, much more than response. It would be hard to use.

@cloudxxx8 , yea I was thinking that would be an issue, but thought that the factory methods could be simple, i.e. no parameters and set ApiVersion, requestID, etc, and then dev has to set the rest of the properties as is done now. Thoughts?

lenny-intel: > I agree with setting ApiVersion as a required field,

@cloudxxx8 , if ApiVersion is required, then it should be set already otherwise it fails validation. So no need to use reflection to make sure it is set.

cloudxxx8: @lenny-intel the client library here is the caller, so it won't validate before sending out the request.  This func can help make sure that every request send out from the client library with the correct api version.

> 
> @cloudxxx8 , yea I was thinking that would be an issue, but thought that the factory methods could be simple, i.e. no parameters and set ApiVersion(s), IDs, etc, and then dev has to set the rest of the properties as is done now. Thoughts?

Could you please write a simple example?

lenny-intel: @cloudxxx8 , @weichou1229 , I have created factory methods for `BaseRequest`, `AddEventRequest`, `Event` and `Reading` in this branch: 
https://github.com/edgexfoundry/go-mod-core-contracts/compare/master...lenny-intel:event-reading-factories

This example does have parameters for most properties and shows how some properties, such as `ApiVersion`,  can be auto filled and others left to be filled in later, such as `Readings`.
I still need to add some unit tests.
If you like this approach I will finish the unit tests and submit a PR.

For the other request factory methods, `Name` should be a parameter and the rest can be left to fill in later. Just depends on the request and its sub-dtos on what works best.

lenny-intel: @cloudxxx8 , @weichou1229 , I have created a PR for my added factory methods for AddEventRequest/Event/Reading DTOs.
https://github.com/edgexfoundry/go-mod-core-contracts/pull/468

Note that I think the factory methods for the other Request and sub DTOs  may be simpler and leave setting bulk of the properties after New() has returned the DTO.

weichou1229: > func NewDeviceRequest (deviceName string, profileName string, serviceName string, adminState string, operatingState string)  AddDeviceRequest 

I thought the method NewDeviceRequest is too many parameters. Can we use few parameters like NewDeviceResponse?
```
func NewAddDeviceRequest(device dtos.Device)(AddDeviceRequest,errors.EdgeX){
	req := AddDeviceRequest{
		BaseRequest: common.NewBaseRequest(),
		Device:       device,
	}

	err := req.Validate()
	if err != nil {
		return req, errors.NewCommonEdgeXWrapper(err)
	}
	return req, nil
}
```

cloudxxx8: No, @weichou1229 ,  Lenny means users can only provide the `name` field

```
func NewAddDeviceRequest(name string) AddDeviceRequest {
	return AddDeviceRequest{
		BaseRequest: common.NewBaseRequest(),
		Device:       dtos.NewDevice(name),
	}
}

func NewDevice(name string) Device {
	return Device{
		Versionable: common.NewVersionable(),
		Id:                uuid.NewString(),
                Name:          name,
	}
}
```

Then, users still have to fill other fields before sending out.

```
 request.Device.xxx= "xxx"
 request.Device.yyy= "yyy"
```

cloudxxx8:
@weichou1229 please close this PR and create a new one for making the factory methods.
Let's start from Device related dtos

PR #486, opened 2021-02-04
feat(v2): Update DeviceServiceCommandClient interface to have queryParams

### LINKED ISSUE #485: [V2] Update DeviceServiceCommandClient interface to have queryParams
The current `DeviceServiceCommandClient` interface only specify `pushEvent` and `returnEvent` as part of func `GetCommand` signature, which is not properly handling user specified query parameters.  The `DeviceServiceCommandClient` interface shall update its `GetCommand` signature to have `queryParams` so that all query params could be passed to device service for further processing.

judehung:
## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features)

## Issue Number:
fix #485 

## What is the new behavior?
The current DeviceServicCommandClient interface only specify pushEvent and returnEvent as part of func GetCommand signature, which is not properly handling user specified query parameters. The DeviceServicCommandClient interface shall update its GetCommand signature to have queryParams so that all query params could be passed to device service for further processing.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

### DIFF
--- a/v2/clients/http/deviceservicecommand.go
+++ b/v2/clients/http/deviceservicecommand.go
@@ -25,13 +25,14 @@ func NewDeviceServiceCommandClient() interfaces.DeviceServiceCommandClient {
 	return &deviceServiceCommandClient{}
 }
 
-func (client *deviceServiceCommandClient) GetCommand(ctx context.Context, baseUrl string, deviceName string, commandName string, pushEvent string, returnEvent string) (responses.EventResponse, errors.EdgeX) {
+func (client *deviceServiceCommandClient) GetCommand(ctx context.Context, baseUrl string, deviceName string, commandName string, queryParams string) (responses.EventResponse, errors.EdgeX) {
 	var response responses.EventResponse
 	requestPath := path.Join(v2.ApiDeviceRoute, v2.Name, url.QueryEscape(deviceName), url.QueryEscape(commandName))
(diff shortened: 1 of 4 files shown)


cloudxxx8 (changes requested):
Please do the same for the Set command.  It might be useful in the future.

judehung:
> Please do the same for the Set command. It might be useful in the future.

@cloudxxx8 for PUT request, the device sdk only consumed BODY and our common client library also don't deal with query strings, do we really need to do this for Set command?

cloudxxx8:
> > Please do the same for the Set command. It might be useful in the future.
> 
> @cloudxxx8 for PUT request, the device sdk only consumed BODY and our common client library also doesn't deal with query strings, do we really need to do this for Set command?

That's for now.  We might use query parameters in the future, so I thought Command Service should pass all parameters anyway.

judehung:
Okay, although it's still awkward to me for a Put request to consume both Body and query strings, https://github.com/edgexfoundry/go-mod-core-contracts/pull/486/commits/c9200ddc3879e5410f65c374f071ff81d8945aac updates the `SetCommand` func with queryParams as parameters.

cloudxxx8 (approved):
LGTM

lenny-intel (approved):
LGTM

PR #525, opened 2021-02-26
feat: Create a common Address struct for V2 API

weichou1229:
## PR Checklist

## What is the new behavior?
Create a common Addressable struct for different endpoint protocols

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information


v2/dtos/addressable.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type RESTAddressable struct {
+	BaseAddressable `json:",inline"`
+	Path            string `json:"path,omitempty"`
+	QueryParameters string `json:"queryParameters,omitempty"`
+	HTTPMethod      string `json:"httpMethod" validate:"required,oneof='GET' 'HEAD' 'POST' 'PUT' 'DELETE' 'TRACE' 'CONNECT'"`
+}
+
+type MQTTAddressable struct {
```
lenny-intel, line 30: Are we assuming only MQTT Publish with this addressable? If so, would the name be better as `MqttPubAddressable`

v2/dtos/addressable.go
```diff
@@ -0,0 +1,85 @@
...
+	BaseAddressable `json:",inline"`
+	Publisher       string `json:"publisher"`
+	Topic           string `json:"topic"`
+	QoS             int    `json:"qos,omitempty"`
```
lenny-intel, line 34: There are other MQTT options that we may need (ignore secure connection for now). They are:
        KeepAlive    // # Seconds (must be 2 or greater)
        Retained     // Boolean
        AutoReconnect  // Boolean
        ConnectTimeout = // Seconds

v2/dtos/addressable.go
weichou1229: > Are we assuming only MQTT Publish with this addressable?

Yes, modified the name.

v2/dtos/addressable.go
weichou1229: Added.

v2/dtos/addressable.go
```diff
@@ -0,0 +1,98 @@
...
+	common.Versionable `json:",inline"`
+
+	RESTAddressable    `json:",inline"`
+	MqttPubAddressable `json:",inline"`
+}
+
+type BaseAddressable struct {
+	Type string `json:"type" validate:"oneof='REST' 'MQTT'"`
+
+	Host string `json:"host"`
```
lenny-intel, line 19: Should also include `Protocol` here.

lenny-intel (changes requested):

v2/dtos/addressable.go
cloudxxx8: "type" here should be able to map "protocol" constants automatically.

lenny-intel: Not 100% since MQTT can be TCP or TCPS, and REST can be HTTP or HTTPS. I am fine if you want to use a different property like `UseTls` or `EnabledTls` or something rather than `Protocol`

cloudxxx8: Since we don't support TLS in the initial support-notifications and support-scheduler, may we just use "MQTT" and "REST" for now?  In the future, we can add "MQTT-TLS" and "REST-TLS" options to the type.  The new `EnableTLS` and other security related fields, such as `secretPath`, can be added in the future, too.
I thought `Protocol` is redundant.

lenny-intel: OK, I see you point. :-)

lenny-intel (approved):
LGTM

cloudxxx8:
from @tonyespy : let’s name it Address,  so as not to confuse it with the old addressable object.
@weichou1229 please go a head to finish this PR with the new naming

weichou1229:
I did the following changes, please help review.
- Rename Addressable to Address
- Fix v2/dtos/addressable.go:44: ambiguous selector a.Type
  - Move Type  from BaseAddress DTO to Address DTO 
- Move Host, Port from BaseAddress DTO to Address DTO and remove BaseAddress DTO because the JSON unmarshal can't identify repeat fields
- Add Validate function to verify the RESTAddress and MqttPubAddress according to the Type.
- Add test for json unmarshalling and validation

lenny-intel (approved):
LGTM

weichou1229:
Renamed MqttPubAddress to MQTTPubAddress.

cloudxxx8 (approved):
LGTM

PR #540, opened 2021-03-10
feat: Implement the re-designed device profile model

### LINKED ISSUE #539: [V2] Implement the re-designed device profile model
- coreCommands is dropped from the device profile

- An attribute (isHidden: true/false with false the default) will be added to the deviceResource and deviceCommand elements to indicate the visibility of any resource or command via Core Command (effectively replacing the coreCommands section of the profile).

- On deviceCommands, we are removing the get/set section.

- On deviceCommands, we are renaming parameter to defaultValue

- On deviceCommands, we will use single resourceOperations to specify selected deviceResources on deviceCommands.

- On deviceCommands, add a **readWrite** indicator to show whether get or set for the deviceCommand is allowed.  The readWrite property can be more restrictive than the associated readWrite property for the deviceResource but it cannot reverse or relax the restriction of the deviceResource (example, make a deviceCommand ‘R’ when the deviceResource is ‘W’ or make the deviceCommand ‘RW’ when the deviceResource is ‘W’). From a schema perspective, the readWrite indicator will **align to what is already the readWrite property on deviceResource**.

- A sourceName will be added to Event DTO to specify the deviceResource or deviceCommand.  The sourceName property will also be used in AutoEvents.

weichou1229:
Close #539

## PR Checklist

## What is the new behavior?
Implement the re-designed device profile model

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information


lenny-intel (dismissed):
LGTM

cloudxxx8 (changes requested):
add omitempty to IsHidden for both json and yaml

weichou1229:
Added the omitempty to IsHidden.

cloudxxx8 (approved):
LGTM

PR #591, opened 2021-04-22
fix: Client lib checks whether the http response is empty

### LINKED ISSUE #590: [V2] Http client lib should check whether the http response is empty
Since we have core-command V2 API  which can execute Get command with **ds-returnevent** parameter, so the DS will return **empty content** when ds-returnevent set to **no**. 

And the empty content will cause json unmarshal error, so the Http client lib should check whether the http response is empty.

https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-command/2.x#/default/get_device_name__name___command_

weichou1229:
Close #590

Signed-off-by: weichou <weichou1229@gmail.com>

## PR Checklist

## What is the new behavior?
Since we have core-command V2 API which can execute Get command with ds-returnevent parameter, so the DS will return empty content when ds-returnevent set to no.
And the empty content will cause json unmarshal error, so the Http client lib should check whether the http response is empty.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information


v2/clients/http/utils/request.go
```diff
@@ -76,7 +80,7 @@ func PostRequestWithRawData(
 	return nil
 }
 
-// Helper method to make the put JSON request and return the body
+// PutRequest make the put JSON request and return the body
```
cloudxxx8, line 83: makes

cloudxxx8 (changes requested):
LGTM

lenny-intel (dismissed):
LGTM, once address @cloudxxx8 comment

v2/clients/http/utils/request.go
weichou1229: Modified.

lenny-intel (approved):
LGTM

cloudxxx8 (approved):
LGTM

PR #593, opened 2021-04-23
fix: Modify DS Get cmd client API to use pointer as return value

### LINKED ISSUE #590: [V2] Http client lib should check whether the http response is empty

weichou1229:
Close #590

## PR Checklist

## What is the new behavior?
Modify DS Get cmd client API to use the pointer as return type. Then we can use the nil pointer to indicate that the HTTP response content is empty.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information


cloudxxx8 (changes requested):
Get Command might return CBOR.
Please check the response header and decide how to deal with it.

v2/clients/http/utils/request.go
```diff
@@ -35,6 +37,36 @@ func GetRequest(ctx context.Context, returnValuePointer interface{}, baseUrl str
 	return nil
 }
 
+// GetRequestAndReturnBinaryRes makes the get request and return the binary response
```
cloudxxx8, line 40: the go doc should state all the returned value, include the contentType

cloudxxx8 (changes requested):

v2/clients/http/utils/request.go
weichou1229: Modified.

cloudxxx8 (approved):
LGTM

PR #613, opened 2021-05-12
refactor: Remove edgex-prefix from all service keys

lenny-intel:
## PR Checklist

Service key constants have the `edgex-` prefix

## What is the new behavior?
Service key constants no longer have the `edgex-` prefix

- [x] Yes
- [ ] No

BREAKING CHANGE: Service key names have changed.

## New Imports

- [ ] Yes
- [x] No

## Specific Instructions

## Other information

### DIFF
--- a/clients/constants.go
+++ b/clients/constants.go
@@ -56,20 +56,19 @@ const (
 
 // Constants related to how services identify themselves in the Service Registry
 const (
-	ServiceKeyPrefix                    = "edgex-"
-	ConfigSeedServiceKey                = "edgex-config-seed"
-	CoreCommandServiceKey               = "edgex-core-command"
-	CoreDataServiceKey                  = "edgex-core-data"
-	CoreMetaDataServiceKey              = "edgex-core-metadata"
-	SupportLoggingServiceKey            = "edgex-support-logging"
-	SupportNotificationsServiceKey      = "edgex-support-notifications"
-	SystemManagementAgentServiceKey     = "edgex-sys-mgmt-agent"
-	SupportSchedulerServiceKey          = "edgex-support-scheduler"
-	SecuritySecretStoreSetupServiceKey  = "edgex-security-secretstore-setup"
-	SecuritySecretsSetupServiceKey      = "edgex-security-secrets-setup"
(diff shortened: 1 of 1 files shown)


lenny-intel:
In draft until reading to push change through all services

clients/constants.go
```diff
@@ -56,20 +56,19 @@ const (
...
-	SecurityBootstrapRedisKey           = "edgex-security-bootstrap-redis"
+	CoreCommandServiceKey               = "core-command"
+	CoreDataServiceKey                  = "core-data"
+	CoreMetaDataServiceKey              = "core-metadata"
+	SupportLoggingServiceKey            = "support-logging"
+	SupportNotificationsServiceKey      = "support-notifications"
+	SystemManagementAgentServiceKey     = "sys-mgmt-agent"
+	SupportSchedulerServiceKey          = "support-scheduler"
+	SecuritySecretStoreSetupServiceKey  = "security-secretstore-setup"
+	SecuritySecretsSetupServiceKey      = "security-secrets-setup"
```
jim-wang-intel, line 67: this one is already gone. should remove it.

jim-wang-intel (changes requested):
This will impact the vault's service key and consul's service key based mechanism. so prepare to change that part too.

jim-wang-intel:
Side note: Although this PR is not related to the service key change of consul, the current Consul's ACL policy rules definition is dependent on the node-name for the instance of `edgex-core-consul`  we are running: for example, the edgexPolicyRules is relying on node_prefix to be "edgex" so that it can confine the access of Consul from the outside world.  For that part, my suggestion is still to retain the "edgex-" prefix for Consul so that only Consul agent node edgex-xxx has written permission.  See details here: https://github.com/edgexfoundry/edgex-go/blob/master/internal/security/bootstrapper/command/setupacl/aclpolicies.go#L45-L47

clients/constants.go
lenny-intel: Done

jim-wang-intel (approved):

iain-anderson (approved):

cloudxxx8 (approved):
LGTM

PR #680, opened 2021-10-21
feat: Add Client API to Pass Get Command Parameters in the Payload

### LINKED ISSUE #677: Add Client API to allow Get Command to pass parameters in the payload
Since we need to pass the parameters in the payload from core-command to Device Services, so we should add new client APIs to pass the payload in the HTTP Get request.

The requirement can refer to https://github.com/edgexfoundry/edgex-go/issues/3754

weichou1229:
Since we need to pass the parameters in the payload from core-command to Device Services, so we add new client APIs to pass the payload in the HTTP Get request.

Closes #677

## PR Checklist

- [x] I have opened a PR for the related docs change (if not, why?)
  https://github.com/edgexfoundry/edgex-go/...

## Testing Instructions

1. Run core-service and device-simple
2. Execute Get command API with payload
3. Debug the protocol driver should receive the payload which is the attributes of the commandRequest.

https://github.com/edgexfoundry/device-sdk-go/blob/1be87f15213c05ab1957194b5dea87ff4eff74a4/pkg/models/commandrequest.go#L15

For example, the payload is:
```
{
    "foo":"123",
    "bar":{
        "foobar":"456"
    }
}
```
The commandRequest should be
```
{
	DeviceResourceName: "testResource"
	Attributes :  {
                "foo":"123",
                "bar":{
                   "foobar":"456"
                           }
               }
	Type : "..."
}
```

### DIFF
--- a/clients/interfaces/deviceservicecommand.go
+++ b/clients/interfaces/deviceservicecommand.go
@@ -17,6 +17,8 @@ import (
 type DeviceServiceCommandClient interface {
 	// GetCommand invokes device service's command API for issuing get(read) command
 	GetCommand(ctx context.Context, baseUrl string, deviceName string, commandName string, queryParams string) (*responses.EventResponse, errors.EdgeX)
+	// GetCommandWithObject invokes device service's get command API and pass the parameters in the payload
+	GetCommandWithObject(ctx context.Context, baseUrl string, deviceName string, commandName string, queryParams string, parameters map[string]interface{}) (*responses.EventResponse, errors.EdgeX)
 	// SetCommand invokes device service's command API for issuing set(write) command
 	SetCommand(ctx context.Context, baseUrl string, deviceName string, commandName string, queryParams string, settings map[string]string) (common.BaseResponse, errors.EdgeX)
 	// SetCommandWithObject invokes device service's set command API and the settings supports object value type
--- a/clients/http/command.go
+++ b/clients/http/command.go
@@ -67,6 +67,19 @@ func (client *CommandClient) IssueGetCommandByName(ctx context.Context, deviceNa
 	return res, nil
 }
 
+// IssueGetCommandByNameWithObject issues the specified read command and pass the parameters in the payload
+func (client *CommandClient) IssueGetCommandByNameWithObject(ctx context.Context, deviceName string, commandName string, dsPushEvent string, dsReturnEvent string, parameters map[string]interface{}) (res *responses.EventResponse, err errors.EdgeX) {
+	requestParams := url.Values{}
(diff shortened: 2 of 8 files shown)


clients/interfaces/deviceservicecommand.go
```diff
@@ -17,6 +17,8 @@ import (
 type DeviceServiceCommandClient interface {
 	// GetCommand invokes device service's command API for issuing get(read) command
 	GetCommand(ctx context.Context, baseUrl string, deviceName string, commandName string, queryParams string) (*responses.EventResponse, errors.EdgeX)
+	// GetCommandWithObject invokes device service's get command API and pass the parameters in the payload
+	GetCommandWithObject(ctx context.Context, baseUrl string, deviceName string, commandName string, queryParams string, parameters map[string]interface{}) (*responses.EventResponse, errors.EdgeX)
```
lenny-intel, line 21: Why can't parameters just be `parameters interface{}` and marshal it to JSON for the body

lenny-intel (changes requested):

clients/interfaces/deviceservicecommand.go
weichou1229: Because I leverage the device-skd-go function to prase the request body as `map[string]interface{}`, see the changes:
https://github.com/edgexfoundry/device-sdk-go/pull/1043/files#diff-7143a51f279090b40470ef1214d5ac141c070ff0b0b1275863da6aca8c4da0abL41-R45

And merge the map to request attributes, see the changes:
https://github.com/edgexfoundry/device-sdk-go/pull/1043/files#diff-54ddce28dd2ff1bcd97f51b3bf766407fb0ac86b85861313fbb37a8cfb459b39R140-R206

So the protocol driver can retrieve the parameter from `reqs []sdkModels.CommandRequest)` in `HandleReadCommands` function.
https://github.com/edgexfoundry/device-sdk-go/blob/1be87f15213c05ab1957194b5dea87ff4eff74a4/example/driver/simpledriver.go#L139

For example, the Get Command request contains the payload:
```
curl --location --request GET 'http://0.0.0.0:59882/api/v2/device/name/Simple-Device01/Switch' \
--header 'Content-Type: application/json' \
--data-raw ' {
    "foo":"123",
    "bar":{
        "foobar":"456"
    },
    "anyDataType": ... 
}'
```
The commandRequest should be
```
{
	DeviceResourceName: "..."
	Attributes :  {
                "foo":"123",
                "bar":{
                   "foobar":"456"
                 },
                 "anyDataType": ...
              }
	Type : "..."
}
```

PR #683, opened 2021-11-05
fix: Fix wrong intervalAction field assignment

### LINKED ISSUE #682: bug: error in ReplaceIntervalActionModelFieldsWithDTO results in nil pointer dereference
The line

https://github.com/edgexfoundry/go-mod-core-contracts/blob/e94f1692727c854a734a8c2863c9766fb8b66556/dtos/requests/intervalaction.go#L121

```
	if patch.ContentType != nil {
		action.ContentType = *patch.Content
	}
```

should be

```
	if patch.ContentType != nil {
		action.ContentType = *patch.ContentType
	}
```

This means that the ContentType can not be updated and also if ContentType is set but Content is nil, then this will result in

```
http: panic serving 127.0.0.1:34958: runtime error: invalid memory address or nil pointer dereference
```

weichou1229:
Closes #682

## PR Checklist

- [ ] I have opened a PR for the related docs change (if not, why?) **not impact the doc**
  <link to docs PR>

## Testing Instructions

Run the unit test to check whether the value is changed as expected.

### DIFF
--- a/dtos/requests/const_test.go
+++ b/dtos/requests/const_test.go
@@ -47,6 +47,7 @@ const (
 	TestPublisher          = "testPublisher"
 	TestTarget             = "testTarget"
 	TestTopic              = "testTopic"
+	TestContent            = "test content"
 
 	TestServiceName = "TestService"
 	TestActionName  = "stop"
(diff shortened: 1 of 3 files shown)


cloudxxx8 (approved):
LGTM

PR #695, opened 2022-01-25
feat: Add ServiceName to Common DTOs

### LINKED ISSUE #696: Add ServiceName to PingResponse
For easier debugging at command line.

AlexCuse:
Uses new method to optionally create with name.

Fixes #696 

Signed-off-by: Alex Ullrich <alex.ullrich@gmail.com>

## PR Checklist

     unit test sufficient
     not sure if docs change needed until used in SDK(s)

### DIFF
--- a/dtos/common/config.go
+++ b/dtos/common/config.go
@@ -12,12 +12,14 @@ package common
 type ConfigResponse struct {
 	Versionable `json:",inline"`
 	Config      interface{} `json:"config"`
+	ServiceName string      `json:"serviceName"`
 }
 
 // NewConfigResponse creates new ConfigResponse with all fields set appropriately
-func NewConfigResponse(serviceConfig interface{}) ConfigResponse {
+func NewConfigResponse(serviceConfig interface{}, serviceName string) ConfigResponse {
 	return ConfigResponse{
 		Versionable: NewVersionable(),
 		Config:      serviceConfig,
+		ServiceName: serviceName,
 	}
 }
--- a/dtos/common/ping.go
+++ b/dtos/common/ping.go
@@ -16,12 +16,14 @@ import (
 type PingResponse struct {
 	Versionable `json:",inline"`
 	Timestamp   string `json:"timestamp"`
+	ServiceName string `json:"serviceName"`
 }
 
 // NewPingResponse creates new PingResponse with all fields set appropriately
-func NewPingResponse() PingResponse {
+func NewPingResponse(serviceName string) PingResponse {
 	return PingResponse{
 		Versionable: NewVersionable(),
 		Timestamp:   time.Now().Format(time.UnixDate),
+		ServiceName: serviceName,
 	}
 }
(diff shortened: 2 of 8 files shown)


dtos/common/ping.go
```diff
@@ -25,3 +26,11 @@ func NewPingResponse() PingResponse {
 		Timestamp:   time.Now().Format(time.UnixDate),
 	}
 }
+
```
lenny-intel, line 29: This code is only used by our services and SDKs. I think we can just add the service name parameter to the existing NewPingResponse(). i.e. it is not a breaking change since the code that uses it is internal. @cloudxxx8 do you agree?

cloudxxx8: It is fine with me, but don't you think we should use Version API instead of Ping?

lenny-intel: @cloudxxx8 , good point. We could add it to both or all the common DTOs (Ping, Version, Metrics and Config) , thoughts?

cloudxxx8: sounds good to me

AlexCuse: This sounds good @lenny-intel @cloudxxx8 will update PR this afternoon

lenny-intel (approved):
LGTM

dtos/common/config.go
```diff
@@ -12,12 +12,14 @@ package common
 type ConfigResponse struct {
 	Versionable `json:",inline"`
 	Config      interface{} `json:"config"`
+	ServiceName string      `json:"serviceName"`
 }
 
 // NewConfigResponse creates new ConfigResponse with all fields set appropriately
-func NewConfigResponse(serviceConfig interface{}) ConfigResponse {
+func NewConfigResponse(serviceConfig interface{}, serviceName string) ConfigResponse {
```
cloudxxx8, line 19: @lenny-intel @AlexCuse 
this is a breaking change
we should revert this PR
a new factory method should be added for this, such as NewXXXResponseWithServiceName
Or, the service name should be inserted to the DTO after it is created

AlexCuse: Thanks @cloudxxx8 new factory method is easy enough - where is the breaking change?

cloudxxx8: @lenny-intel said it has been considered, so I am fine with it.
Basically, go-mod-core-contracts is a public module and NewConfigResponse is a public function.
If there is someone use NewConfigResponse in go-mod-core-contracts in their code, the compile error will happen once they upgrade the go-mod-core-contracts version

AlexCuse: Ah got it thanks @cloudxxx8 - my initial PR did have both but we removed on scope expansion to keep it simple.  Thanks!

lenny-intel: @cloudxxx8, @AlexCuse, I  Brought this up at today's TSC and we are ok with this breaking change since this code is meant to only be used by our services and SDKs. Risk of breaking someone else code is very low.

PR #699, opened 2022-02-16
feat: add ProtocolPropertiesRequest for validate API

hahattan:
Signed-off-by: Chris Hung <chris@iotechsys.com>

## PR Checklist

- [ ] I have opened a PR for the related docs change (if not, why?) N/A
  <link to docs PR>

### DIFF
--- a/dtos/requests/intervalaction.go
+++ b/dtos/requests/intervalaction.go
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2021 IOTech Ltd
+// Copyright (C) 2021-2022 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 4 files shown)


lenny-intel:
@hahattan , why is this needed? I thought that the new device would be sent to the Validate endpoint which would validate the ProtocolProperties

hahattan:
> @hahattan , why is this needed? I thought that the new device DTO would be sent to the Validate endpoint which would validate the ProtocolProperties

I thought we're only sending the `Protocols` field in device DTO .
This is unnecessary if we're sending device DTO, if that's the case we can close this PR and I'll make change in SDK interface.

lenny-intel:
> I thought we're only sending the `Protocols` field in device DTO .

Seems sending the Device DTO simplifies it so we don't have to add this code and opens it up to other validation beyond protocols if ever needed. @cloudxxx8 thoughts?

cloudxxx8:
@hahattan @lenny-intel since the Device DTO is relative small, it would be fine send out the whole Device DTO to validate.
On the other hand, the Device can log the device name or other information for debugging.

@hahattan however, it's good you catch some typo.  Could you please open another PR for the go doc typo and unit test correction?  thanks.

hahattan:
last question: 
per the discussion in https://github.com/edgexfoundry/device-sdk-go/pull/1097#issuecomment-1040587647
I think I should use `AddDeviceRequest` right ? (although the naming is confusing in SDK side)
https://github.com/edgexfoundry/go-mod-core-contracts/blob/58b8b4bc41ce6f7a1dbe3f3412e4301586b7a828/dtos/requests/device.go#L21-L24

cloudxxx8:
Yes, let's reuse it to simplify the contract

cloudxxx8:
If this happens in Update calls and it contains the request contains protocolProperties changes, we will use UpdateDeviceRequest to avoid confusion.  It should contain the full protocolProperties anyway. 
https://github.com/edgexfoundry/go-mod-core-contracts/blob/58b8b4bc41ce6f7a1dbe3f3412e4301586b7a828/dtos/requests/device.go#L63-L66

PR #727, opened 2022-04-20
feat: add new method for CommandClient

### LINKED ISSUE #713: Command Client missing capability to passes additional query parameters for Get command
IssueGetCommandByName() doesn't have a parameter for addition query parameters.

Need to add new method `IssueGetCommandByNameWithQueryParms` to allows the user query parameters to be added to the Get command request.

hahattan:
add 'IssueGetCommandByNameWithQueryParams' method to allow the user
query parameters to be added to the Get command request.

fixes #713

Signed-off-by: Chris Hung <chris@iotechsys.com>

## PR Checklist

  <link to docs PR>

## Testing Instructions

1. use `CommandClient` from this branch and verify the new method is working

### DIFF
--- a/clients/http/command.go
+++ b/clients/http/command.go
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2021 IOTech Ltd
+// Copyright (C) 2021-2022 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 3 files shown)


clients/http/command_test.go
```diff
@@ -50,6 +50,18 @@ func TestIssueGetCommandByName(t *testing.T) {
...
 
+func TestIssueGetCommandByNameWithQueryParams(t *testing.T) {
+	deviceName := "Simple-Device01"
+	cmdName := "SwitchButton"
+	path := path.Join(common.ApiDeviceRoute, common.Name, deviceName, cmdName)
+	ts := newTestServer(http.MethodGet, path, &responses.EventResponse{})
+	defer ts.Close()
+	client := NewCommandClient(ts.URL)
+	res, err := client.IssueGetCommandByNameWithQueryParams(context.Background(), deviceName, cmdName, map[string]string{"foo": "bar"})
+	require.NoError(t, err)
```
lenny-intel, line 61: Can you add code to verify the query params where actually in the resulting REST request.

clients/http/command.go
```diff
@@ -67,6 +67,20 @@ func (client *CommandClient) IssueGetCommandByName(ctx context.Context, deviceNa
...
 }
 
+func (client *CommandClient) IssueGetCommandByNameWithQueryParams(ctx context.Context, deviceName string, commandName string, queryParams map[string]string) (res *responses.EventResponse, err errors.EdgeX) {
+	requestParams := url.Values{}
+	for k, v := range queryParams {
+		requestParams.Set(k, v)
+	}
+
+	requestPath := path.Join(common.ApiDeviceRoute, common.Name, url.QueryEscape(deviceName), url.QueryEscape(commandName))
+	err = utils.GetRequest(ctx, &res, client.baseUrl, requestPath, requestParams)
```
lenny-intel, line 77: Is it ok that the `dsPushEvent`  and `dsReturnEvent` parameters are not sent. I.e. does the Device Service have a default for these if not present? If not, we need to add defaults here is they are not present in the passed in `queryParams`

lenny-intel (changes requested):

clients/http/command_test.go
cloudxxx8: @hahattan  we have the default values in the core-command, so we should just add the mock function to behave as the core-command default values here

clients/http/command.go
cloudxxx8: @lenny-intel since we have the default value in core-command, I don't think we should add the default value in the client library to prevent the double maintenance effort if we want to change the default values in the future.

hahattan: Device service also have default value:
https://docs.edgexfoundry.org/2.0/design/adr/device-service/0011-DeviceService-Rest-API/#query-parameters

clients/http/command_test.go
lenny-intel: Need to get this merged ASAP, once TSC approved after code Freeze

lenny-intel (approved):
LGTM

cloudxxx8 (approved):
LGTM

PR #761, opened 2022-09-28
feat: Reduce the character restriction of the name field

### LINKED ISSUE #750: Reduce the character restriction of the `name` field
# 🚀 Feature Request

 This feature request is for all services

Driven by https://github.com/edgexfoundry/edgex-go/issues/4053

We added `ValidateDtoRFC3986UnreservedChars` func to check if DTO's name pointer value only contains unreserved characters as defined in https://tools.ietf.org/html/rfc3986#section-2.3
unreserved characters= ALPHA / DIGIT / "-" / "." / "_" / "~"
Also due to names used in topics for Redis Pub/Sub, "." is not allowed

However, we noticed colon `:` is a standard character of BACnet device resource.  It might look like `abc:def` in the resource name.
Furthermore, a customer told us their device might contain space and dollar sign in the resource name.
Thus, I propose to remove this restriction.

Users can still use URL escape to use those characters to call the API.
Our go-mod-messaging might also help escape those characters

Remove `edgex-dto-rfc3986-unreserved-chars` validation tag from all the `name` fields.

weichou1229:
Add characters :;= to reduce the character restriction.
Add characters :;= to reduce the character restriction.
- In BACNet protocol, the user might combine object type and property as the resourceName, for example, analog_input_0:present-value
- In OPC_UA protocol, the user might use NodeId as the resourceName, for example, ns=10;s=Hello:World

Close #750

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) not impact the test
- [ ] I have opened a PR for the related docs change (if not, why?) not impact the doc
  <link to docs PR>

## Testing Instructions

- Run core-metadata
- Add metadata with characters :;=

### DIFF
--- a/common/validator.go
+++ b/common/validator.go
@@ -34,7 +34,7 @@ const (
 const (
 	// Per https://tools.ietf.org/html/rfc3986#section-2.3, unreserved characters= ALPHA / DIGIT / "-" / "." / "_" / "~"
 	// Also due to names used in topics for Redis Pub/Sub, "."are not allowed
-	rFC3986UnreservedCharsRegexString = "^[a-zA-Z0-9-_~]+$"
+	rFC3986UnreservedCharsRegexString = "^[a-zA-Z0-9-_~:;=]+$"
 	intervalDatetimeLayout            = "20060102T150405"
 	name                              = "Name"
 )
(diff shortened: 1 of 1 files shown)


lenny-intel:
@weichou1229  Why `;` & `=` ? Issue #750 only mentions `:`

weichou1229:
> @weichou1229 Why `;` & `=` ? Issue #750 only mentions `:`

Sorry, added the explanation to the comment.
```
- In BACNet protocol, the user might combine object type and property as the resourceName, for example, analog_input_0:present-value
- In OPC_UA protocol, the user might use NodeId as the resourceName, for example, ns=10;s=Hello:World
```

lenny-intel (approved):
LGTM

lenny-intel:
@weichou1229 do we have this is the documentation and if we do it needs to be updated. If not please add an issue to document this character restriction on names.

weichou1229:
No problem, create the issue https://github.com/edgexfoundry/edgex-docs/issues/872

PR #782, opened 2022-12-23
refactor!: Use true/false for command parameters to be more consistent

### LINKED ISSUE #780: [EdgeX 3.0] Use true/false instead of yes/no in query parameters
# 🚀 Feature Request

 This feature request is for core-command

We are using common.ValueYes and common.ValueNo for some query parameters, and we should use true and false for consistency.

Remove common.ValueYes and common.ValueNo, and change the code to use common.ValueTrue and common.ValueFalse

jinlinGuan:
BREAKING CHANGE:
- ds-pushevent and ds-returnevent to use true/false instead of yes/no

close #780

Signed-off-by: Ginny Guan <ginny@iotechsys.com>

## PR Checklist

- [x] I have opened a PR for the related docs change (if not, why?)
https://github.com/edgexfoundry/edgex-docs/pull/917

## Testing Instructions

build core-command and device-virtual docker images based on the changes and run docker-compose.yml
The following commands work as expected:
- `curl http://localhost:59882/api/v2/device/name/Random-Boolean-Device/Bool\?ds-pushevent\=true`
- `curl http://localhost:59882/api/v2/device/name/Random-Boolean-Device/Bool\?ds-pushevent\=false`
- `curl http://localhost:59882/api/v2/device/name/Random-Boolean-Device/Bool\?ds-return\=true`
- `curl http://localhost:59882/api/v2/device/name/Random-Boolean-Device/Bool\?ds-return\=false`

### DIFF
--- a/clients/interfaces/command.go
+++ b/clients/interfaces/command.go
@@ -23,9 +23,9 @@ type CommandClient interface {
 	// DeviceCoreCommandsByDeviceName returns all commands associated with the specified device name.
 	DeviceCoreCommandsByDeviceName(ctx context.Context, deviceName string) (responses.DeviceCoreCommandResponse, errors.EdgeX)
 	// IssueGetCommandByName issues the specified read command referenced by the command name to the device/sensor that is also referenced by name.
-	// dsPushEvent: If set to yes, a successful GET will result in an event being pushed to the EdgeX system. Default value is no.
-	// dsReturnEvent: If set to no, there will be no Event returned in the http response. Default value is yes.
-	IssueGetCommandByName(ctx context.Context, deviceName string, commandName string, dsPushEvent string, dsReturnEvent string) (*responses.EventResponse, errors.EdgeX)
+	// dsPushEvent: If set to true, a successful GET will result in an event being pushed to the EdgeX system. Default value is false.
+	// dsReturnEvent: If set to false, there will be no Event returned in the http response. Default value is true.
+	IssueGetCommandByName(ctx context.Context, deviceName string, commandName string, dsPushEvent bool, dsReturnEvent bool) (*responses.EventResponse, errors.EdgeX)
 	// IssueGetCommandByNameWithQueryParams issues the specified read command by deviceName and commandName with additional query parameters.
 	IssueGetCommandByNameWithQueryParams(ctx context.Context, deviceName string, commandName string, queryParams map[string]string) (*responses.EventResponse, errors.EdgeX)
 	// IssueSetCommandByName issues the specified write command referenced by the command name to the device/sensor that is also referenced by name.
--- a/common/constants.go
+++ b/common/constants.go
@@ -190,8 +190,6 @@ const (
 	DefaultOffset  = 0
 	DefaultLimit   = 20
 	CommaSeparator = ","
-	ValueYes       = "yes"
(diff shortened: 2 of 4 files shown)


clients/interfaces/command.go
```diff
@@ -23,8 +23,8 @@ type CommandClient interface {
 	// DeviceCoreCommandsByDeviceName returns all commands associated with the specified device name.
 	DeviceCoreCommandsByDeviceName(ctx context.Context, deviceName string) (responses.DeviceCoreCommandResponse, errors.EdgeX)
 	// IssueGetCommandByName issues the specified read command referenced by the command name to the device/sensor that is also referenced by name.
-	// dsPushEvent: If set to yes, a successful GET will result in an event being pushed to the EdgeX system. Default value is no.
-	// dsReturnEvent: If set to no, there will be no Event returned in the http response. Default value is yes.
+	// dsPushEvent: If set to true, a successful GET will result in an event being pushed to the EdgeX system. Default value is false.
+	// dsReturnEvent: If set to false, there will be no Event returned in the http response. Default value is true.
 	IssueGetCommandByName(ctx context.Context, deviceName string, commandName string, dsPushEvent string, dsReturnEvent string) (*responses.EventResponse, errors.EdgeX)
```
lenny-intel, line 28: Can we make this interface use `bool` instead of `string` for these options? 
```suggestion
	IssueGetCommandByName(ctx context.Context, deviceName string, commandName string, dsPushEvent bool, dsReturnEvent bool) (*responses.EventResponse, errors.EdgeX)
```

common/constants.go
```diff
@@ -190,8 +190,6 @@ const (
 	DefaultOffset  = 0
 	DefaultLimit   = 20
 	CommaSeparator = ","
-	ValueYes       = "yes"
-	ValueNo        = "no"
 	ValueTrue      = "true"
 	ValueFalse     = "false"
```
lenny-intel, line 194: Do we really need these constants for `true and `false`??
```suggestion
	ValueTrue      = true
	ValueFalse     = false
```

lenny-intel (changes requested):

lenny-intel:
Please change commit message to have breaking change all on one line. The Conventional Commit spec what it all one one line 2 blank line???.  I think 2 blank line is used when there is a very long description. I missed this on the other PR :-( .
https://github.com/edgexfoundry/go-mod-core-contracts/blob/main/.github/Contributing.md#footer

```
BREAKING CHANGE: ds-pushevent and ds-returnevent to use true/false instead of yes/no
```

common/constants.go
cloudxxx8: yes, we still need them for the query parameter definition

cloudxxx8: they should still be in string data type, because the REST query parameter only allows string value

lenny-intel: Ok, can we then parse them to bool so that the `IssueGetCommandByName` func can use bool?

cloudxxx8: Sure, `IssueGetCommandByName` client can use bool directly, and `ValueTrue`/`ValueFalse` will be used in the core-command internal implementation.

lenny-intel (approved):
LGTM

cloudxxx8 (approved):
LGTM

Issue #772, opened 2022-11-30
[EdgeX 3.0] Provision Watcher model improvement


cloudxxx8:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for core-metadata

### Description [**REQUIRED**]
This is issue is opened for discussing the Provision Watcher model improvment in v3.0.

### Describe the solution you'd like
The first suggestion is to modify type of the `Identifiers` to `map[string][]string`, which is the same as the `BlockingIdentifiers`

cloudxxx8:
Another suggestion is to add an extendable field in this object:
`Properties map[string]interface`
For example, the Device Service can support `DeviceNameTemplate` to allow users to define the generated device name.

judehung:
> ### Describe the solution you'd like
> The first suggestion is to modify type of the `Identifiers` to `map[string][]string`, which is the same as the `BlockingIdentifiers`

For the type change on PW's Identifiers, we will address it in #592.  For this #772, we will simply add an extendable field `Properties map[string]any` into ProvisionWatcher dto/model.

PR #803, opened 2023-02-17
feat: Add new Properties field to ProvisionWatcher DTO and Model

judehung:
Add an extendable field into ProvisionWatcher DTO and Model: `Properties map[string]any`. This is useful when certain device services requires extra information from ProvisionWatcher.  For example, assume a device service would like to generate the device name in certain format during auto discovery, the user can define a property "DeviceNameTemplate" with value as the template of device name in the ProvisionWatcher, so that the device service can pick up such property to generate the device name.

fixes https://github.com/edgexfoundry/go-mod-core-contracts/issues/772

## PR Checklist

  <link to docs PR>


cloudxxx8 (approved):
LGTM

PR #806, opened 2023-02-20
feat: Use URL escape for device command name and resource name

### LINKED ISSUE #773: Use URL escape for device command name and resource name
# 🚀 Feature Request

 This feature request is for core-metadata

Related to https://github.com/edgexfoundry/go-mod-core-contracts/issues/750
Some protocol would use `/` or `.` for the device or resource name.  By using the URL escape, device command name and resource name can allow lots of characters.

weichou1229:
Remove the constraint of device command name and resource name, and use URL escape in client API.

Close #773

## PR Checklist

- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

## Testing Instructions

- Run core service and device service
- Verify the escape URL works, eg, escape the commandName `device-a/test:value` and put into the URL like  /api/v2/device/name/Modbus-TCP-Device/device-a%2Ftest%3Avalue

### DIFF
--- a/clients/http/utils/common.go
+++ b/clients/http/utils/common.go
@@ -1,5 +1,5 @@
 //
-// Copyright (C) 2020-2022 IOTech Ltd
+// Copyright (C) 2020-2023 IOTech Ltd
 // Copyright (C) 2023 Intel Corporation
 //
 // SPDX-License-Identifier: Apache-2.0
@@ -75,11 +75,10 @@ func makeRequest(req *http.Request, authInjector interfaces.AuthenticationInject
 }
 
 func createRequest(ctx context.Context, httpMethod string, baseUrl string, requestPath string, requestParams url.Values) (*http.Request, errors.EdgeX) {
-	u, err := url.Parse(baseUrl)
+	u, err := url.Parse(baseUrl + requestPath)
 	if err != nil {
 		return nil, errors.NewCommonEdgeX(errors.KindServerError, "fail to parse baseUrl", err)
 	}
-	u.Path = path.Join(u.Path, requestPath)
 	if requestParams != nil {
 		u.RawQuery = requestParams.Encode()
 	}
@@ -92,11 +91,10 @@ func createRequest(ctx context.Context, httpMethod string, baseUrl string, reque
 }
 
 func createRequestWithRawDataAndParams(ctx context.Context, httpMethod string, baseUrl string, requestPath string, requestParams url.Values, data interface{}) (*http.Request, errors.EdgeX) {
-	u, err := url.Parse(baseUrl)
+	u, err := url.Parse(baseUrl + requestPath)
 	if err != nil {
 		return nil, errors.NewCommonEdgeX(errors.KindServerError, "fail to parse baseUrl", err)
 	}
-	u.Path = path.Join(u.Path, requestPath)
 	if requestParams != nil {
 		u.RawQuery = requestParams.Encode()
 	}
@@ -120,11 +118,10 @@ func createRequestWithRawDataAndParams(ctx context.Context, httpMethod string, b
 }
 
 func createRequestWithRawData(ctx context.Context, httpMethod string, baseUrl string, requestPath string, requestParams url.Values, data interface{}) (*http.Request, errors.EdgeX) {
-	u, err := url.Parse(baseUrl)
+	u, err := url.Parse(baseUrl + requestPath)
 	if err != nil {
 		return nil, errors.NewCommonEdgeX(errors.KindServerError, "fail to parse baseUrl", err)
 	}
-	u.Path = path.Join(u.Path, requestPath)
 	if requestParams != nil {
 		u.RawQuery = requestParams.Encode()
 	}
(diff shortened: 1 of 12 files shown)


clients/http/reading_test.go
```diff
@@ -141,7 +141,7 @@ func TestQueryReadingsByDeviceNameAndResourceNamesAndTimeRange(t *testing.T) {
 	defer ts.Close()
 
 	client := NewReadingClient(ts.URL)
-	res, err := client.ReadingsByDeviceNameAndResourceNamesAndTimeRange(context.Background(), deviceName, resourceNames, start, end, 1, 10)
+	res, err := client.ReadingsByDeviceNameAndResourceNamesAndTimeRange(context.Background(), deviceName, resourceNames, start, end, start, end)
```
lenny-intel, line 144: Why was this change as part of this PR?

clients/http/utils/common.go
```diff
@@ -67,11 +67,10 @@ func makeRequest(req *http.Request) (*http.Response, errors.EdgeX) {
 }
 
 func createRequest(ctx context.Context, httpMethod string, baseUrl string, requestPath string, requestParams url.Values) (*http.Request, errors.EdgeX) {
-	u, err := url.Parse(baseUrl)
+	u, err := url.Parse(baseUrl + requestPath)
```
lenny-intel, line 70: `+` is not equivalent to the `path.Join` previously.  Same change needed in all locations

Also why is this change part of this PR? Has nothing to do with URL escaping?

```suggestion
	u, err := url.Parse(path.Join(u.Path, requestPath))
```

dtos/corecommand.go
```diff
@@ -16,7 +16,7 @@ type DeviceCoreCommand struct {
 // CoreCommand and its properties are defined in the APIv2 specification:
 // https://app.swaggerhub.com/apis-docs/EdgeXFoundry1/core-command/2.1.0#/CoreCommand
 type CoreCommand struct {
-	Name       string                 `json:"name" validate:"required,edgex-dto-none-empty-string,edgex-dto-rfc3986-unreserved-chars"`
+	Name       string                 `json:"name" validate:"required,edgex-dto-none-empty-strin"`
```
lenny-intel, line 19:
```suggestion
	Name       string                 `json:"name" validate:"required,edgex-dto-none-empty-string"`
```

lenny-intel (changes requested):

clients/http/reading_test.go
weichou1229: Right, it's a redundant change, will revert.

dtos/corecommand.go
weichou1229: Thank you for catching that.

clients/http/utils/common.go
weichou1229: We use u.String() to create an HTTP request object
`req, err := http.NewRequest(httpMethod, u.String(), nil)`

But the previous request URL is not expected, it does escape again:
```
func TestUrlStringFunction(t *testing.T) {
	baseUrl := "http://localhost:59980"
	requestPath := "/api/v2/device/name/Modbus-TCP-Device/device-a%2Ftest%3Avalue"

	u1, _ := url.Parse(baseUrl)
	u1.Path = path.Join(u1.Path, requestPath)
	fmt.Println("1.", u1.String())

	u2, _ := url.Parse(baseUrl + requestPath)
	fmt.Println("2.", u2.String())

	u3, _ := url.Parse(path.Join(baseUrl, requestPath))
	fmt.Println("3.", u3.String())
}
```
```
1. http://localhost:59980/api/v2/device/name/Modbus-TCP-Device/device-a%252Ftest%253Avalue
2. http://localhost:59980/api/v2/device/name/Modbus-TCP-Device/device-a%2Ftest%3Avalue
3. http:/localhost:59980/api/v2/device/name/Modbus-TCP-Device/device-a%2Ftest%3Avalue
```

lenny-intel: Also, `path.Join()` is for file paths and will generate the path differently when running on windows. Not sure it is appropriate here for a URL path.

lenny-intel: Neverminded ;-)
```
The path package should only be used for paths separated by forward slashes, such as the paths in URLs. This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the path/filepath package.
```

lenny-intel (approved):
LGTM

cloudxxx8 (changes requested):
this should also be handled
https://github.com/edgexfoundry/go-mod-core-contracts/blob/1751cc4a89d3720a418614bd06f95962415fbbdc/clients/http/event.go#L39

weichou1229:
> this should also be handled
> 
> https://github.com/edgexfoundry/go-mod-core-contracts/blob/1751cc4a89d3720a418614bd06f95962415fbbdc/clients/http/event.go#L39

Updated.

cloudxxx8 (approved):
LGTM

clients/http/utils/common.go
ajcasagrande: https://go.dev/play/p/f3Ig2PcArH3

ajcasagrande:
```go
func createRequestWithEncodedData(ctx context.Context, httpMethod string, baseUrl string, requestPath string, data []byte, encoding string) (*http.Request, errors.EdgeX) {
	u, err := parseBaseUrlAndRequestPath(baseUrl, requestPath)
	if err != nil {
		return nil, errors.NewCommonEdgeX(errors.KindServerError, "failed to parse baseUrl and requestPath", err)
	}
```

```go
func parseBaseUrlAndRequestPath(baseUrl, requestPath string) (*url.URL, error) {
	fullPath, err := url.JoinPath(baseUrl, requestPath)
	if err != nil {
		return nil, err
	}
	return url.Parse(fullPath)
}
```

cloudxxx8: @ajcasagrande is there an issue or are you suggesting a refactoring?

cloudxxx8: oh, I saw the issue opened.  please ignore my question.

PR #809, opened 2023-02-22
feat!: update data type in ResourceProperties

### LINKED ISSUE #770: [EdgeX 3.0] Device Profile model improvement
# 🚀 Feature Request

 This feature request is for core-metadata

This is issue is opened for discussing the Device Profile model improvment in v3.0.

The first suggestion is that Value Properties base, offset, scale etc become numerics rather than strings

hahattan:
BREAKING CHANGE: update mask,shift,base,scale,offset to numeric data type

also rename 'Others' to 'Optional' for the consistency with message broker settings.

closes #770

## PR Checklist

  <link to docs PR>

### DIFF
--- a/dtos/deviceprofile_test.go
+++ b/dtos/deviceprofile_test.go
@@ -23,6 +23,7 @@ var testAttributes = map[string]interface{}{
 }
 var testMappings = map[string]string{"0": "off", "1": "on"}
 var testTags = map[string]any{"TestTagsKey": "TestTagsValue"}
+var testOptional = map[string]any{"isVirtual": false}
 
 var testDeviceProfile = models.DeviceProfile{
 	Name:         TestDeviceProfileName,
@@ -37,7 +38,7 @@ var testDeviceProfile = models.DeviceProfile{
 		Properties: models.ResourceProperties{
 			ValueType: common.ValueTypeInt16,
 			ReadWrite: common.ReadWrite_RW,
-			Others:    testTags,
+			Optional:  testOptional,
 		},
 		Tags: testTags,
 	}},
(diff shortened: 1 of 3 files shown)


dtos/deviceprofile_test.go
```diff
@@ -37,7 +37,7 @@ var testDeviceProfile = models.DeviceProfile{
 		Properties: models.ResourceProperties{
 			ValueType: common.ValueTypeInt16,
 			ReadWrite: common.ReadWrite_RW,
-			Others:    testTags,
+			Optional:  testTags,
```
lenny-intel, line 40: Why using `testTags`? Optional is not for tags, so please use something more relevant to what we expect it to be used for.

lenny-intel (changes requested):

lenny-intel (approved):
LGTM

iain-anderson (changes requested):
Can we use json.Number for scale, offset and base - so that they can be integer or float?

hahattan:
> Can we use json.Number for scale, offset and base - so that they can be integer or float?

`json.Number` is alias of `string` which provides useful methods to transform value to numeric data type.
But there still will be error in the same place if user provides illegal value:
https://github.com/edgexfoundry/device-sdk-go/blob/b46b419fd49731d0deb4deed8e6127ef6a42900c/internal/transformer/transformresult.go#L92-L93

Using `float64` directly can avoid the error during device profile creation but it's a breaking change.

Another way to improve this is to use `json.Number` for  these properties, but leverage validator to check the type.
For example https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Numeric for `base`, `scale` and `offset`.
But we might need to implement custom tag for `mask` and `shift`

@lenny-intel @cloudxxx8 @iain-anderson thought ?

cloudxxx8:
@iain-anderson is it ok?

iain-anderson:
OK with this as is

PR #811, opened 2023-03-02
feat: Specify how to authenticate outbound IntervalActions

bnevis-i:
Outbound HTTP requests must be authenticated if they call EdgeX services (for example, event aging).  This commit extends the model to allow users to specify the AuthMethod of outbound HTTP requests.

## PR Checklist

- [X] I am not introducing a new dependency (add notes below if you are)
  <link to docs PR>

### DIFF
--- a/dtos/intervalaction.go
+++ b/dtos/intervalaction.go
@@ -20,6 +20,7 @@ type IntervalAction struct {
 	Content      string  `json:"content,omitempty"`
 	ContentType  string  `json:"contentType,omitempty"`
 	AdminState   string  `json:"adminState" validate:"oneof='LOCKED' 'UNLOCKED'"`
+	AuthMethod   string  `json:"authMethod" validate:"oneof='' 'NONE' 'JWT'"`
 }
 
 // NewIntervalAction creates intervalAction DTO with required fields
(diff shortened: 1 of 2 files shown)


dtos/intervalaction.go
```diff
@@ -20,6 +20,7 @@ type IntervalAction struct {
 	Content      string  `json:"content,omitempty"`
 	ContentType  string  `json:"contentType,omitempty"`
 	AdminState   string  `json:"adminState" validate:"oneof='LOCKED' 'UNLOCKED'"`
+	AuthMethod   string  `json:"authMethod" validate:"oneof='' 'none' 'jwt'"`
```
lenny-intel, line 23: Value should be be all caps as in line above for consistency. I assume the empty value will work, but haven't seen it used prior. 

```suggestion
	AuthMethod   string  `json:"authMethod" validate:"oneof='' 'NONE' 'JWT'"`
```

lenny-intel (changes requested):

dtos/intervalaction.go
bnevis-i: Changed to upper case.

dtos/intervalaction.go
```diff
@@ -42,6 +43,7 @@ type UpdateIntervalAction struct {
 	ContentType  *string  `json:"contentType"`
 	Address      *Address `json:"address"`
 	AdminState   *string  `json:"adminState" validate:"omitempty,oneof='LOCKED' 'UNLOCKED'"`
+	AuthMethod   *string  `json:"authMethod" validate:"omitempty,oneof='' 'none' 'jwt'"`
```
lenny-intel, line 46: Upper case needed here also

lenny-intel (changes requested):

dtos/intervalaction.go
bnevis-i: @lenny-intel Sorry 'bout that.

lenny-intel (approved):
LGTM

PR #846, opened 2023-08-02
feat: Add new API route constants for Echo

### LINKED ISSUE #845: Add new API route constants with path params for Echo
# 🚀 Feature Request

This feature request is for all of the EdgeX micro services have a dependency on it.

The original `gorilla/mux` package uses the following format to define the URL path with parameters:
```
/api/v3/device/name/{name}
```

However, the `Echo` [package](https://echo.labstack.com/)(the replacement of `gorilla/mux`) uses a different format as below:
```
/api/v3/device/name/:name
```

New constants are added in `echo_api_constants.go` to define the new API route constants with URL path parameters.

lindseysimple:
Closes #845. Add new API route constants with path params for Echo.

## PR Checklist

  <link to docs PR>

### DIFF
--- a/common/echo_api_constants.go
+++ b/common/echo_api_constants.go
@@ -1,0 +1,82 @@
+//
+// Copyright (C) 2023 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+// Constants related to defined routes with path params in the v3 service APIs for Echo
+const (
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

lenny-intel:
Why did we create new constants rather than simply adjusting the existing ones?
Using the `echo` package name in the constant is not good practice.

cloudxxx8:
The original ones are public constants, and they may be used by other projects with gorilla, such as eKuiper or any rest clients.
It's breaking changes.
The idea is to use the original constants name in v4 and remove those temporary echo ones.
I didn't notice the path param declaration is quite different.

lenny-intel:
OK, so please add comment that these will be removed in v4

PR #855, opened 2023-09-06
feat!: Allow NameFieldEscape configurable

weichou1229:
BREAKING CHANGE: Allow NameFieldEscape configurable and add API path builder for Http and MQTT client

## PR Checklist

- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

## Testing Instructions

- Run unit test
- Related PRs are opened, you can refer to https://github.com/edgexfoundry/edgex-go/issues/4673#issuecomment-1709699336 and run services to verify the API.


common/utils.go
```diff
@@ -52,3 +52,35 @@ func URLEncode(s string) string {
...
+type pathBuilder struct {
+	sb                    strings.Builder
+	enableNameFieldEscape bool
+}
+
+func NewPathBuilder() *pathBuilder {
+	return &pathBuilder{}
+}
+
+func (b *pathBuilder) EnableNameFieldEscape(enableNameFieldEscape bool) *pathBuilder {
```
lenny-intel, line 65: Why not pass `enableNameFieldEscape bool` to `NewPathBuilder` and eliminate this function?

weichou1229: I thought it was not a required value for the pathBuilder, so I didn't put it into the construct function.

lenny-intel: > I thought it was not a required value for the pathBuilder, so I didn't put it into the construct function.

Looks like you are always calling EnableNameFieldEscape so seems it is required.

`common.NewPathBuilder().EnableNameFieldEscape`

weichou1229: Yes, I am always calling the function,  but we will remove it in EdgeX 4.0, so would it be optional in the future?

lenny-intel: It will be removed, so it will not exist for being optional. Also the clean-up for 4.0 will be easier (slightly) if it is a parameter to the NewPathBuilder. ;-) This not a big deal, so approving PR as is since device services are failing again without it.

lenny-intel (approved):

PR #887, opened 2024-04-04
feat: Add optional "parent" field to Device objects.

eaton-coreymutter:
Related to issue edgexfoundry/edgex-go#4769 but does not close it, other PRs to come.

## PR Checklist

- [ x ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [ x ] I am not introducing a new dependency (add notes below if you are)
- [ x ] I have added unit tests for the new feature or bug fix (if not, why?)
- [ x ] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
  Docs to update are in edgex-go (the OpenAPIs), will update those in another PR for this issue

## Testing Instructions

I built core-metadata, core-command, device-virtual with these changes, verified I could set the "parent" field via PATCH or via inclusion in YAML device definition file, the field can be set to empty via PATCH with empty-string value, and the field persists across restarts of core-metadata.


dtos/device.go
```diff
@@ -45,8 +47,10 @@ type UpdateDevice struct {
 // ToDeviceModel transforms the Device DTO to the Device Model
 func ToDeviceModel(dto Device) models.Device {
 	var d models.Device
+	d.DBTimestamp = models.DBTimestamp(dto.DBTimestamp)
```
eaton-coreymutter, line 50: Noticed this was missing when adding the new unit test.

eaton-coreymutter: ... and it occurs to me that we probably don't want DBTimestamp to propagate here... will update.

common/constants.go
```diff
@@ -191,6 +191,7 @@ const (
 	PushEvent    = "ds-pushevent"   //query string to specify if an event should be pushed to the EdgeX system
 	ReturnEvent  = "ds-returnevent" //query string to specify if an event should be returned from device service
 	RegexCommand = "ds-regexcmd"    //query string to specify if the command name is in regular expression format
+	Parent       = "descendantsOf"  //Limit returned devices to those who have parent, grandparent, etc. of the given device name
```
cloudxxx8, line 194: since there is no namespace for the constant, we need to make the constant name understandable.
please rename to `DecendantsOf` or `ParentQueryParam`
according to other naming convention, I would suggest using `DecendantsOf`

cloudxxx8 (changes requested):

common/constants.go
eaton-coreymutter: Done - also my team pointed out a maxLevels query parameter would be useful, OK to add that at this late stage?

cloudxxx8: Yeah, I agree with this, please add it

AnilChoudhury-Eaton (commented):
Looks good.

cloudxxx8 (approved):
LGTM

cloudxxx8:
@eaton-coreymutter the device client mock should be regenerated, or we will get the unit test failure in other services
https://github.com/edgexfoundry/device-sdk-go/pull/1575#pullrequestreview-2016210517

cloudxxx8:
[2024-04-23T04:16:06.573Z] --- FAIL: TestAddReadingTags (0.00s)
[2024-04-23T04:16:06.573Z] panic: interface conversion: *mocks.DeviceClient is not interfaces.DeviceClient: missing method AllDevicesWithChildren [recovered]
[2024-04-23T04:16:06.573Z] 	panic: interface conversion: *mocks.DeviceClient is not interfaces.DeviceClient: missing method AllDevicesWithChildren

fixed in https://github.com/edgexfoundry/go-mod-core-contracts/pull/889

eaton-coreymutter:
> @eaton-coreymutter the device client mock should be regenerated, or we will get the unit test failure in other services [edgexfoundry/device-sdk-go#1575 (review)](https://github.com/edgexfoundry/device-sdk-go/pull/1575#pullrequestreview-2016210517)

Sorry about that - I regenerated it once but forgot to do it again after changing the signature.

cloudxxx8:
no problem

PR #934, opened 2024-10-07
fix: Added validation for UpdateScheduleJobRequest

jackchenjc:
refer to https://github.com/edgexfoundry/edgex-go/issues/4925

## PR Checklist

  <link to docs PR>

### DIFF
--- a/dtos/requests/schedulejob.go
+++ b/dtos/requests/schedulejob.go
@@ -61,10 +61,26 @@ type UpdateScheduleJobRequest struct {
 }
 
 // Validate satisfies the Validator interface
-func (a *UpdateScheduleJobRequest) Validate() error {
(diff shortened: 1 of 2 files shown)


cloudxxx8 (changes requested):
use dive or inline validation tag in dto, and the validation can be called automatically

jackchenjc:
Since there is no `Validation` func for `UpdateScheduleJob` dto, the validation should be called in the validation function of `UpdateSubscriptionRequest`.

If we create a `Validation` func for `UpdateScheduleJob` dto, the validation can be called automatically, but that function will be very similar to the one for `ScheduleJob` dto.

cloudxxx8 (approved):
LGTM

PR #935, opened 2024-10-14
refactor: Using isNull instead of nil reading value

weichou1229:
Using isNull to indicate the reading value should be nul after marshaling.

## PR Checklist

  <link to docs PR>

### DIFF
--- a/dtos/reading.go
+++ b/dtos/reading.go
@@ -32,10 +32,11 @@ type BaseReading struct {
 	BinaryReading `json:",inline" validate:"-"`
 	SimpleReading `json:",inline" validate:"-"`
 	ObjectReading `json:",inline" validate:"-"`
+	NullReading   `json:",inline" validate:"-"`
 }
 
 type SimpleReading struct {
-	Value *string `json:"value"`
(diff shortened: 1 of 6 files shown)


dtos/reading.go
```diff
@@ -35,7 +35,8 @@ type BaseReading struct {
 }
 
 type SimpleReading struct {
-	Value *string `json:"value"`
+	IsNull bool   `json:"isNull,omitempty"` // indicate the reading value should be null in the JSON payload
```
cloudxxx8, line 38:
```suggestion
	isNull bool   `json:"isNull,omitempty"` // indicate the reading value should be null in the JSON payload
```
don't expose this field

is there any difficulty to move this field to a new NullReading struct?

weichou1229: Updated and created the NullReading struct.
Is the NullReading only used for simple reading?

cloudxxx8: No, it can be applied to any valueType
it will cause the value, binaryValue, and objectValue all `: null`

cloudxxx8 (approved):
LGTM

Issue #989, opened 2025-03-08
Precision limited on Float64 values: Issue or Design?


jrtitus:
Given the following code:

https://github.com/edgexfoundry/go-mod-core-contracts/blob/db8453dd8a72e3689edf5a1f241eef43b810904a/dtos/reading.go#L201-L207

A Float64 resource's value will be truncated and rounded at the 6th decimal point:

```go
value := 1.234567890987654321
fmt.Printf("%e", value)

// Output:
//   1.234568e+00
```

A solution might be:

```go
func convertFloatValue(valueType string, kind reflect.Kind, value any) (string, error) { 
 	if err := validateType(valueType, kind, value); err != nil { 
 		return "", err 
 	} 

        if kind == reflect.Float64 {
                return fmt.Sprintf("%.15e", value), nil
        }
  
 	return fmt.Sprintf("%e", value), nil 
 } 
```

```go
value := 1.234567890987654321
fmt.Printf("%.15e", value)

// Output:
//   1.234567890987654e+00
```

Is this something you see as an issue, or is it more of a design decision? Thanks for your time.

cloudxxx8:
this is an issue, not a design decision, so we should fix it

PR #995, opened 2025-03-10
fix: parse float to string with precision to a sufficient level

judehung:
fixes https://github.com/edgexfoundry/go-mod-core-contracts/issues/989

The original implemetation uses `fmt.Sprintf("%e", value)` to convert float to string, which will be truncated and rounded at the 6th decimal point. For a float value that needs more precision (not arbitrary precision), it's better to use `strconv.FormatFloat` with prec -1 to achieve the smallest number of digits necessary such that `ParseFloat` will return f exactly. Moreover, as `fmt.Sprintf` is for more general purpose and `strconv.FormatFloat` is specifically for formatting a float, the `strconv.FormatFloat` has better performance than `fmt.Sprintf`.

The benchmark test shows the `strconv.FormatFloat` takes 955.5 ns/op and `fmt.Sprintf` takes 1961 ns/op.

## PR Checklist

  <link to docs PR>

### DIFF
--- a/dtos/event_test.go
+++ b/dtos/event_test.go
@@ -147,7 +147,7 @@ func TestEvent_AddSimpleReading(t *testing.T) {
 		value        string
 	}{
 		{int32(12345), "myInt32", common.ValueTypeInt32, "12345"},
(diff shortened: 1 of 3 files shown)


cloudxxx8 (approved):
LGTM

PR #997, opened 2025-03-18
build(deps): revise go directive to 1.23.0

judehung:
Per https://github.com/dependabot/dependabot-core/issues/11825 and https://github.com/dependabot/dependabot-core/pull/11826 ,the dependabot has been updated to use GOTOOLCHAIN=go1.24.1 and it will add toolchain in `go mod tidy` when the go directive is specified with language version (1.N). As we don't want the toolchain to be specified with go1.24.1, change the go directive to released version (1.N.P).

## PR Checklist

- [ ] I am not introducing a new dependency (add notes below if you are)
  <link to docs PR>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/edgexfoundry/go-mod-core-contracts/v4
 
-go 1.23
+go 1.23.0
 
 require (
 	github.com/fxamacker/cbor/v2 v2.7.0


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/edgexfoundry/go-mod-core-contracts/v4
 
-go 1.23
+go 1.23.0
+
+toolchain go1.23.7
```
cloudxxx8, line 5: do you know if this toolchain impact anything?

judehung: The toolchain will specify the version of standard library as well as the compiler, assembler, and other tools. I suspect that dependbot fails due to missing toolchain, so just gave it a try. As I'm still investigating this issue, convert this PR to draft.

judehung: Well, I pushed another commit https://github.com/edgexfoundry/go-mod-core-contracts/pull/997/commits/4a47f5b9bc59228376409a9280fda300c610fb5a to remove the toolchain in go.mod, and the `make test` fails at the same error as dependabot.

judehung:
@cloudxxx8 Per https://github.com/dependabot/dependabot-core/issues/11825 and https://github.com/dependabot/dependabot-core/pull/11826 , the dependabot has been updated to use GOTOOLCHAIN=go1.24.1 and it will add toolchain in go mod tidy when the go directive is specified with language version(1.N). As we don't want to the toolchain to be specified with go1.24.1, change the go directive to released version(1.N.P).

cloudxxx8 (approved):
LGTM

PR #1012, opened 2025-08-07
feat: Reduce payload size for device events

weichou1229:
The proposal introduces a new global environment variable, EDGEX_OPTIMIZE_EVENT_PAYLOAD, for all EdgeX services. When EDGEX_OPTIMIZE_EVENT_PAYLOAD is set to true, the event payload will be optimized.

## PR Checklist

  <link to docs PR>

## Testing Instructions

- Deploy core service, device-virtual, app service
- Inspect the event message to see whether the payload is reduced.

### DIFF
--- a/dtos/event.go
+++ b/dtos/event.go
@@ -1,35 +1,39 @@
 //
-// Copyright (C) 2020-2023 IOTech Ltd
+// Copyright (C) 2020-2025 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
 package dtos
 
 import (
+	"encoding/json"
 	"encoding/xml"
+	"os"
 	"time"
 
-	"github.com/edgexfoundry/go-mod-core-contracts/v4/dtos/common"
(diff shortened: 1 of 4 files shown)


dtos/event.go
```diff
@@ -96,3 +100,101 @@ func (e *Event) ToXML() (string, error) {
...
+		aux.Readings = make([]BaseReading, len(e.Readings))
+	}
+
+	if os.Getenv(common.EnvOptimizeEventPayload) == common.ValueTrue {
+		for i, reading := range e.Readings {
+			reading.Id = ""
+			reading.DeviceName = ""
+			reading.ProfileName = ""
+			if e.Origin == reading.Origin {
+				reading.Origin = 0
```
judehung, line 140: Just to confirm, will a zero origin be stripped out the reading when omitempty is specified?

I think the goal of ADR is to remove origin field out of reading when its value is identical to the event-level origin.

judehung (approved):
LGTM

